PR #30, opened 2026-05-10
feat: Sync Center — Phase 1 (IssueStore + 413 retry-loop fix)

Rasbandit:
## Summary

- New `IssueStore` — path-keyed, persistent map of sync failures classified by category (`too_large` / `auth` / `server` / `network` / `conflict` / `other`).
- `pushFile` catch block routes errors through `categorizeError`: records the issue, **skips the offline queue when the failure is terminal** (e.g. 413), clears the issue on next successful push.
- Persisted alongside `lastSync` / `syncState` so the upcoming Sync Center pane sees a stable list across reloads.

## Why now

Health vault reported "299 notes / 106 attachments to push" but pushAll silently looped — every PDF >5MB hit the backend's 413 cap, the offline queue re-enqueued the same payload, and the next retry hit 413 again. No surface anywhere to learn which file or why. This patch makes failures *visible and terminal* instead of *invisible and infinite*.

## What's in this PR

| Area | Change |
|---|---|
| `src/issue-store.ts` (new) | `IssueStore` class + `categorizeError()` helper |
| `src/types.ts` | `SyncIssue` + `SyncIssueCategory` types |
| `src/sync.ts` | pushFile catch block: record + classify + conditional enqueue; success path clears issue |
| `src/main.ts` | hydrate on load, persist with plugin data |
| `tests/issue-store.test.ts` (new) | 18 unit tests — store dedupe/persist/categorize |
| `tests/sync.test.ts` | 3 integration tests — 413 records + skips queue, 500 records + queues, success clears |

**Tests: 397 pass / 0 fail (was 394).** Build clean (tsc + esbuild + biome).

## What's NOT in this PR (follow-up commits on this branch)

This is Phase 1 of the Sync Center redesign. Subsequent phases land as additional commits on this branch before merge:

| # | Scope |
|---|---|
| 2 | `SyncCenterView` workspace-leaf skeleton + ribbon icon |
| 3 | Issues section live (rendering IssueStore) |
| 4 | Activity section live (replaces SyncLogModal) |
| 5 | Actions strip + inline confirm (replaces PreSyncModal) |
| 6 | Stats + header status (live counts) |
| 7 | Retire PreSyncModal + SyncLogModal + actions in settings tab |

Per project convention (single PR per feature, version bump at PR-open) — version moves `1.2.1 → 1.3.0` once at the top of the branch and the PR re-renders as new phases land.

## Test plan

- [ ] `bun test` — all 397 green
- [ ] `bun run build` — tsc + esbuild + biome clean
- [ ] Manual: push a >5MB attachment → IssueStore captures with `category: too_large`, status 413, sizeBytes set
- [ ] Manual: confirm the offline queue does NOT grow on 413
- [ ] Manual: reload plugin → issues persist
- [ ] Manual: shrink the file + retry → issue clears on success


Rasbandit:
### Code review

Found 1 issue:

1. `categorizeError` marks HTTP 401/403 with `terminal: false`, so the offline queue keeps retrying after a permanent auth failure. `src/api.ts` already does an inline single-shot refresh-and-retry on 401 — if a 401 still bubbles up to `pushFile`'s catch, refresh has already failed (revoked key, account suspended, etc.). Re-enqueueing then loops every flush. 403 has no auto-recovery path at all and should also be terminal. Same loop pattern that prior fixes (e.g. 404-on-delete) explicitly closed.

https://github.com/Rasbandit/engram-obsidian-sync/blob/a9deea5a32956e45a9f635a6c8fc732726a28ef9/src/issue-store.ts#L91-L95

PR #31, opened 2026-05-11
fix: sync preview manifest inventory + consolidate actions in Sync Center

Rasbandit:
## Summary

- **Root-cause fix for bogus sync-preview numbers.** `PreSyncModal` was showing things like "server: 0 notes / local: 299 to push / 104 attachments to push" on the health vault even though everything was already synced. `computeSyncPlan` was using the `/notes/changes?since=lastSync` **delta** as the server inventory — long-synced files don't appear in the delta, so they got wrongly flagged as toPush.
- `computeSyncPlan` now calls `/sync/manifest` for the authoritative path set + `serverNoteCount`. The `/changes` feed is still consulted at `since=lastSync` for content diff (pull/conflict). When the server doesn't expose a manifest, the engine falls back to `since=epoch` so inventory stays correct at the cost of one full-content fetch.
- **Sync / Push all / Pull all actions move into the Sync Center** (sidebar leaf + settings-tab mirror). The Cloud and Self-hosted tabs now show a single "Open Sync Center" CTA. Every action opens `PreSyncModal` before running.
- `plugin.openProgressModal()` exposed so the sidebar leaf can open the progress modal without depending on the settings tab.

Bumps 1.3.0 -> 1.3.1.

## Test plan

- [ ] Reload the health vault and open Sync Center. Preview shows real `serverNoteCount`, empty `toPush.notes`, empty `toPush.attachments` once everything is synced.
- [ ] Sync / Push all / Pull all in Sync Center each open `PreSyncModal` and run the chosen flow with the progress modal.
- [ ] Pull all -> "Wipe & pull" still gates on `WipeConfirmModal`.
- [ ] Cloud + Self-hosted tabs: only the "Open Sync Center" CTA in the Actions row.
- [ ] Manifest-absent fallback: pointing at an older self-hosted server still produces a correct preview (slow but accurate).


Rasbandit:
Closing per audit — this PR ended up mixing 5 unrelated concerns (sync preview manifest fix, OAuth refresh-token persistence, telemetry log, locally-modified detection, Sync Center CTA), and is now 25 commits behind main with 71-file drift. Each commit is being re-extracted onto a fresh branch off main as its own focused PR per the single-PR-per-concern rule.

PR #120, opened 2026-06-19
fix(ws): shorten heartbeat 30s→15s — server idle-close was dropping broadcasts (#646)

Rasbandit:
## Root cause (the e2e-clerk WS-broadcast flake)

Diagnosed via server-side `SyncChannel.terminate` logging on a real failing run: the sync socket closes **server-side** with `{:shutdown, :local_closed}` (34×/run). That's Phoenix's WebSocket **idle `:timeout`** (default 60s) firing.

The plugin heartbeat (`channel.ts` `startHeartbeat`) ran every **30s** — only a 2× margin under the 60s server timeout. Under e2e-clerk's 2-worker load the headless-Obsidian renderer starves the `setInterval` tick; a heartbeat slips past 60s; the server idle-closes the socket; the client reconnects (1s); and any `note_changed` broadcast sent during that gap is **lost** (broadcasts are fire-and-forget to currently-joined sockets). That's the test_47/48/78 flake family — and it's e2e-clerk-specific only because that variant runs 2 workers (more CPU contention than e2e-local's 1).

Not OAuth (these clients authenticate the socket via ApiKeyAuth) and not the #119 pull path (live updates apply directly in `handleStreamEvent`, bypassing `pull()`).

## Fix

Heartbeat 30s → **15s** (extracted to a documented `HEARTBEAT_MS` constant) — a 4× margin under the 60s timeout, so even a badly-delayed tick still beats it. Prod-safe and cheap.

Pairs with backend WS-timeout headroom + the e2e Clerk-leak fix (separate engram PR).

## Tests
`tests/channel.test.ts` 20/20 pass; eslint + biome + tsc/esbuild build clean.

Refs #646.


Rasbandit:
Superseded by #121. The serial+heartbeat verification disproved the load/heartbeat theory — the real root cause was the stale-epoch zombie channel (async connectChannel race on re-auth), fixed in #121 and validated against the 15×-repeat harness. Closing; the 15s heartbeat was harmless but isn't the fix.

PR #153, opened 2026-06-29
test(crdt): live-editor + reconnect-reenroll bundle for e2e [DO NOT MERGE]

Rasbandit:
**Throwaway test bundle — do not merge.** Combines two independent features on top of current \`main\` (which already has #152) so the whole CRDT live-sync experience can be tested as one build + rc:

- the new **live editor binding** (\`feat/crdt-live-editor-binding\`): binds Obsidian's live CodeMirror editor to the per-path Y.Text, so Obsidian edits sync live instead of in ~5s disk-flush bursts. Covers main editor, frontmatter, multi-pane, reading view; closed notes keep the disk path; internal-API hooks fall back to disk.
- **#151** reconnect re-enroll (live sync survives reconnect).

Version is a throwaway **1.9.99** so the RC workflow cuts \`1.9.99-rc.1\` without consuming the real \`1.9.26-rc.*\` line. Each feature ships as its own PR; this branch is test-only.

**Pair with** engram \`feat/web-crdt-editor\` (#784) for the web side (\`make saas-dev BACKEND_DIR=…/engram/.worktrees/feat-web-crdt-editor\`) so the web app is a live CRDT peer.

**Test:** open the same \`.md\` note in Obsidian + web → type in Obsidian, expect live (not bursty); type in web → live in Obsidian; concurrent edits both survive; frontmatter / two-pane / reading-view.

Gate: 1388 tests / 0 fail, tsc + build + lint:obsidian clean.


Rasbandit:
Superseded by #160 — the live-editor binding extracted onto current main (post #151/#154, hardening included). This bundle was test-only by design.

PR #155, opened 2026-07-02
fix(crdt): client hardening wave (audit 2026-07-01, post-1.10)

Rasbandit:
Eight fixes closing the remaining plugin-side findings from the 2026-07-01 CRDT sync audit, on top of 1.10.0 (proto-2 + Y.Map). Each landed TDD-first with a per-task review; the branch passed a final whole-branch review whose two Critical findings are fixed in the last commit.

## ⚠️ Deploy ordering (hard requirement)

**Backend PR engram#846 must be merged AND release-tagged to prod BEFORE this plugin version ships wide.** The new-note design here is legacy-first: an empty/unsynced doc declines CRDT and pushes via REST, relying on #846's convergent merge (tail replay + three-way vs snapshot). Against a pre-#846 backend, declined pushes never reach the server CRDT doc and a later STEP2 can deliver stale doc content over disk.

## Fixes

1. **Handshake-gated seeding** (audit P0-1) — a fresh-IndexedDB device no longer seeds every note as a new Yjs lineage racing the server's STEP2 (vault-wide body duplication). Final semantics after review: a path is marked synced only when an applied inbound frame leaves the doc non-empty; new-note content always travels the legacy path first and the doc adopts the server's merged lineage; marks are cleared on disconnect and join errors.
2. **One-time v1-doc IndexedDB wipe** — pre-1.10 fence-in-body lineages are dropped once per vault (schema marker); docs rebuild from the server; disk keeps un-pushed edits.
3. **Sync gate on all CRDT paths** (audit P0-2) — cold-start reconcile, enrollment (active-leaf + doc-ready), inbound flush, and empty-note materialization all respect the direction-choice gate; gate-accept resets enrollment so handshakes re-fire. Prevents a vault switch from overwriting local files with the other vault's content.
4. **Offline CRDT capture** (audit P1-1) — after the first successful crdt join, disconnects keep CRDT routing: md edits accumulate in the Y.Doc + IndexedDB and travel via the reconnect handshake; the version-less offline-queue snapshot replay is retired for consumed md edits. Teardown no longer depends on a connection-state transition (zombie-manager fix).
5. **Doc teardown on delete/rename** (audit P1-3) — `removeDoc` (in-memory + IndexedDB) wired into local delete, rename (old path), and remote delete; no ghost lineage resurrection on recreate-at-path.
6. **Live-event enroll parity** (audit P2-1) — a `note_changed` for an unobserved CRDT note now enrolls it instead of only logging.
7. **Join-error degrade + proto-too-old notice** (audit F13 + T4 finding) — a failed crdt (re)join now tears down CRDT routing (degrades to the convergent legacy path) instead of silently dropping frames; `crdt_proto_too_old` additionally shows a once-per-session update notice. Join errors are ref-matched so #846's per-message `rate_limited`/`frame_too_large` replies do NOT tear down the session.
8. **Heartbeat reply timeout** (audit P2-2) — an unanswered heartbeat closes the half-dead socket (mobile resume state) so the normal reconnect machinery takes over.

## Verification

- 1460 tests, 0 failures (full suite, three runs across the wave); `tsc` + esbuild + biome + eslint (obsidian rules) + stylelint clean
- Final whole-branch review walked the full lifecycle state matrix (connect/disconnect/rejoin-error/gate/delete/heartbeat × synced/enrolled/initiated/crdtEverJoined) — the two Criticals it found (gated doc-ready frames reverting remote edits on restart; join-ref collision with per-message error replies) are fixed in the final commit with tests

## Dependencies / related work

- **PR #151** (open): re-enroll open notes on crdt (re)join — complements this wave; without it, an offline-edited note that stays the active leaf re-handshakes only on the next enroll trigger (edit, tab switch, pull). This wave deliberately does not duplicate it.
- The live-editor CRDT binding is a separate in-flight branch; untouched here.

## Follow-ups being filed

Join-error degrade with unsent doc ops (dual-lineage risk); flushQueue delete-replay IDB ghost (lineage contamination on recreate); suppressDeletes on remote delete (redundant round-trip, pre-existing); paced pull-discovery enrollment for 1000+-note fresh installs under the 240/10s crdt_msg limit.


Rasbandit:
## Live-sync regression — diagnosis (2026-07-02)

Ran the sync e2e suite locally against a CRDT stack (recipe now in workspace `docs/context/local-crdt-e2e-repro.md`, ~20s/run). Findings:

**This branch regresses live CRDT sync.** Against the fixed backend (post-#880), plugin `main` fails only 1 flaky test; this branch fails ~11 (test_02/03/04/10/13/36/37/38/43/49/50/78 — all live delivery/merge).

**It's the adopt-first fix being defeated, not a conflict-merge issue.**
- `test_37_append_sync`: B ends with the append-only delta, base content GONE. Time-snapshots show B's file materializes already append-only — the base was adopted then lost, and an append-only delta landed on an empty doc (state-vector/doc mismatch).
- This is an OLD bug (branch base #154 fails test_37) that **main FIXED via #161 adopt-first seed gate** (current main PASSES). This branch reintroduces it despite the merge bringing #161 in.

**Sync-gate commit `8f834b0` is EXONERATED** — cherry-picked onto current main, test_37 still passes. The culprit is in the seed/doc-lifecycle commits (prime suspects `0f9155e` retire-legacy-md/offline-capture, `e15daa7` schema-wipe, `5ebb99a` gate-seeding), which are woven through the receive path and not cleanly revertable.

**Not a conflict-fix away from mergeable.** Recommend either (a) instrument the receive path + pinpoint/fix the one commit that defeats adopt-first, or (b) decompose: land the isolated safe commits (delete/rename teardown `c535a97`, half-dead socket close `d1cfcef`, enroll-on-note_changed `c1e7786`) as a small green PR and hold the seed/offline-capture rework for dedicated work.

Merge conflicts with main are already resolved + pushed (a828da3); plugin-side CI is green. The block is the e2e regression above.

PR #178, opened 2026-07-06
fix(sync): don't invalidate a valid token on a network-blip WS close

Rasbandit:
## Problem

Live sync between devices chronically dies and edits stop propagating; recovery required a full Obsidian/tab reload. Diagnosed from prod Loki + Tempo: a burst of WebSocket `auth rejected` / `reason=signature_error` recurring daily.

Root cause is in `src/channel.ts`'s `onclose` handler. It treated **any** fast pre-open `1006` close (within `AUTH_FAIL_WINDOW_MS`, before `onopen`) as a stale-token rejection and called `invalidateAccessToken()`. A network/DNS blip (`UnknownHostException: Unable to resolve host "api.engram.page"`) produces the **identical** bare-`1006` close. So a blip threw away a valid token, the refresh then failed while offline, and a stale token was replayed on reconnect, which the backend rejected as `signature_error`. The live-sync channel never re-established.

This is not the same trigger as #129 (backend switch); this close-heuristic still shipped in 1.11.x.

## Fix

On a fast pre-open close, stop guessing:
- **Offline** (`navigator.onLine === false`) → network drop. Keep the token, back off. Never invalidate.
- **Online** → fire one authenticated `GET /me` probe (`setAuthProbe`) and let `api.ts`'s existing real-`401` handler be the **single** invalidation authority (RFC 6750: only a real `401 invalid_token` proves the token is stale). A `2xx` means the token is fine and the close was transport noise.

The close-code guess is deleted. `onclose` stays **synchronous** and the probe is fire-and-forget (`void this.authProbe().catch(...)`) so it can't defer reconnect scheduling or race an intentional `disconnect()`. The handler also now logs the full `CloseEvent` (code/reason/wasClean) + online state.

## Companion

Backend observability PR (attribution for any residual `signature_error`, plus per-user trace filtering): engram-app/Engram — branch `feat/auth-reject-attribution-trace-attrs`.

## Testing

TDD: fast-close-online → probe fired, token NOT invalidated; offline → no probe, token preserved; rejecting probe → no throw, reconnect still scheduled. `bun test` 1585 pass / 0 fail; build + biome + lint:obsidian + lint:css clean.


Rasbandit:
Companion backend observability PR: engram-app/Engram#921

PR #180, opened 2026-07-06
feat(crdt): key CRDT by note_id — rename-stable live sync

Rasbandit:
**DRAFT — coordinated cutover. Pairs with backend PR engram-app/Engram#925. Merge + deploy together.**

Keys the plugin's CRDT layer by the note's stable `note_id` instead of `{vaultId}/{path}`, so an in-app rename/move keeps the same live CRDT room (fixes `test_34_folder_rename`).

## What changed
- **`NoteIdMap`** (`src/crdt/note-id-map.ts`): a sidecar `path→note_id` map (no frontmatter pollution) with a reverse `pathForId` lookup, persisted in `data.json`. Maintained across Obsidian `rename` events (id stable, key moves).
- **Mint + learn** (`src/sync.ts`): a brand-new note mints a UUIDv7 and sends it as the create id; existing notes learn their `note_id` from the `/sync/changes` pull. Conflict-resolution branches also write the authoritative server id back into the map.
- **CRDT keyed by note_id** (`src/crdt/manager.ts`, `channel.ts`, `main.ts`, `CrdtLiveViews`): local Y.Doc + wire frames keyed by `note_id`; inbound frames resolve `note_id → path` via `pathForId` for disk flush; unknown inbound id is a graceful no-op. Deleted `CrdtManager.renameDoc` (re-keying on rename would destroy live history — the exact bug). `dbPrefix` demoted to IndexedDB-storename only (never the wire id). The `.md` gate stays where the path is known. Reuses the `#915` `crdtLive` gate for routing.
- **Field-name fix**: `pushNote` now sends the client-minted id under the body key **`id`** (backend reads `id`, not `client_id`) — without this the server silently ignored the minted id.

## Verification
Per-task TDD + independent per-task review (the loop caught + fixed the reverse-map test gap, the `rename()` reverse-cleanup asymmetry, and the field-name mismatch). Full suite **1604/0**, biome + eslint + build clean, v1.11.12. **CI on this draft (incl. the cross-repo e2e trigger against backend#925) is the coordinated proof.**

Refs engram-app/Engram#915 (the CRDT-push-join-gate work this builds on) and the id-keyed doc_id spec.


Rasbandit:
Coordinated pair: backend PR is engram-app/Engram#925. Merge + deploy together.

PR #185, opened 2026-07-07
fix(sync): materialize never-seen CRDT notes on live delivery (test_47)

Rasbandit:
## Root cause (e2e test_47 OAuth WebSocket live-sync, ~50% flake)

With the `crdt:` topic joined, an API-created note arrives as a `note_changed` broadcast carrying its id. The C1 "CRDT-managed" skip in `handleStreamEvent` then suppressed the body write for **every** `.md` — including notes this device has **never seen** — and left delivery to the `crdt_doc_ready` announce. That announce is edge-triggered/fire-and-forget; if the client isn't subscribed to `crdt:` when it fires (or the handshake exceeds the timeout), the note **never lands**. The test waits only on the *sync* topic being connected, so it races the `this.crdt` flip. #643 bumped the timeout 10s→15s — a band-aid on delivery *loss*, not latency.

The **pull path** (`applyChange`) already materializes never-seen notes on discovery; the WS path didn't — this aligns them.

## Fix

Narrow the C1 skip in `handleStreamEvent`:
- **Already-local** note → keep skipping the legacy write (CRDT owns live edits; avoids the double-write feedback loop).
- **Never-seen** note → materialize the body now from the broadcast (inline `content`, else fetch via `getNote`) using echo-safe `flushFromCrdt`, instead of relying on the racy announce. A later announce re-flush no-ops (idempotent).

## Tests
- New: never-seen note materializes from the broadcast; already-local note does NOT write.
- Updated the 4 C1-gate tests that pinned the old rely-on-announce behavior (they encoded the bug). Full suite green (1651 pass).

Likely also helps the same CRDT-discovery family (#646 / test_48). e2e verification in progress.


Rasbandit:
Closing: this approach (eager never-seen materialization on the live WS path) fixes test_47 but regresses test_10, test_34, and test_78 (verified — main baseline passes solo, this branch fails all three solo). It races the id-keyed rename/move machinery and clobbers hash-only live updates. Tracking the flake + better directions in a fresh issue; not shipping a 3-test regression for a self-healing prod flake.

PR #190, opened 2026-07-07
test: sync seam coverage — sendCrdt join gate, wiring convergence

Rasbandit:
## Context

Part of the cross-repo testing-confidence wave driven by the workspace audit doc `docs/context/testing-surface-audit-2026-07-07.md` (engram-workspace repo). Companion backend PR to follow.

## Behavior change (prominent)

`sendCrdt` now returns `boolean` and **drops frames until the `crdt:` join is server-acked.** Previously it emitted frames with a stale/absent `join_ref` that the server silently discarded (the #179 failure shape) — so nothing that worked before stops working; the drop is now honest and local. No caller consumes the return yet, by design; content self-heals via STEP1 resync on join. The old test asserting the pre-join emit was deliberately flipped.

## What else is in this PR

- Pure extraction of `main.ts` CRDT wiring into `src/crdt/wiring.ts` (data plane; behavior-preserving, review-verified). One documented deviation: pending stranded flushes are now discarded on channel re-setup (`wiring.dispose()` cancels the heal timer and drops the pending map), whereas the old code kept them at plugin level so a pending strand-heal survived `setupNoteStream()` re-setup; recovery still exists via the new session's per-connect reconcile + re-handshake.
- Two-device convergence tests with **real per-device keying**, replacing the old regex-keyed integration-test approach. These are regression pins for the append-doubling / receiver-move-dup (#182) / push-before-join (#179) bug classes.
- Resumed-device reconcile coverage was audited: already pinned by #187's tests, nothing added here.

## Coverage boundary

The two-device convergence scenario for rename covers the wiring seam (map + disk level), not full `SyncEngine` rename propagation — documented coverage boundary, not a gap being claimed as closed.

## Review-fix included

Deleted the dead write-only `crdtChannel` field in `src/main.ts` (zero reads, verified by review). `crdtManager` and `crdtEnrollment` stay (still read).

## Version

Bumped 1.11.18 → 1.11.19 (main had already advanced to 1.11.18 via #188 after this branch was cut, so the originally-planned 1.11.17 → 1.11.18 bump was stale; re-bumped against current main).

---


Rasbandit:
Merged main (#193, v1.11.20) — clean, no overlap with the wiring extraction. Re-bumped to 1.11.21 (1.11.22 reserved for another in-flight PR). Full gauntlet green (biome, eslint, stylelint, 1676 tests, build).

Rasbandit:
Version-queue note (as of 2026-07-07 evening): main=1.11.20; **#190 and #194 both currently claim 1.11.21**; #196 claims 1.11.22. Whichever of #190/#194 merges second needs a re-bump (and #196 re-bumps if the ladder shifts). Merge order is Todd's call — this is just so nobody is surprised by the version-check.

PR #193, opened 2026-07-07
fix(crdt): cross-wired note ids caused cross-file overwrite

Rasbandit:
## Summary

Fixes a confirmed prod data-loss incident (2026-07-07): file B's content silently overwrote file A on focus-switch. Three files collapsed to one content hash; the overwrite reached the server (not just local), amplified by two live CRDT clients on one vault. Recovery required Obsidian's local File Recovery snapshots since the server copies were poisoned too.

## Root cause

A stale/empty region of the client `noteIdMap` plus `getOrMint` could leave a note_id `X` mapped to path `B` while the server authoritatively assigns `X -> A` and `B -> Y`. `onFlushToDisk(X)` then resolved `B` via `pathForId` and wrote `A`'s live CRDT content onto `B` — cross-wiring two unrelated notes.

Two contributing gaps:
- `NoteIdMap.set()` left an id's OLD path in the forward map when reassigning it, so two paths could resolve to one id and `pathForId` flip-flopped.
- `reconcileNoteIdMapFromManifest()` was fill-unknown only (it skipped any id it already had mapped), so a cross-wire — an id pointing at another note's real path — was never corrected. Fill-unknown was originally chosen to protect in-flight renames (the test_10 guard); that tradeoff was the gap.

## Fix shape (two commits)

1. **`1cde4d5` fix(crdt): correct cross-wired note ids to stop cross-file overwrite** — `NoteIdMap.set()` evicts the stale forward entry on reassignment, keeping the map a clean bijection. `reconcileNoteIdMapFromManifest()` now rebinds an id to the manifest's authoritative path when the local path is itself a manifest path owned by a different id, while still preserving genuine in-flight renames.

2. **`2a3d5eb` fix(crdt): guard destructive ops against a lying noteIdMap** — hardens the destructive relocate path that an independent verification pass showed was still open after commit 1: `moveIfIdRelocated` now checks manifest ownership before trashing a prior path (cross-wire -> rebind, never trash; unknowable -> rebind, skip trash, since a duplicate is recoverable but a wrong trash isn't). `NoteIdMap` also rejects garbage paths (a literal `"null"` key was found minted + CRDT-enrolled in the incident's `data.json`). Cold-start now reconciles from the manifest before the startup mint loop, since a stale map there was the highest-volume wrong-mint site. New replication test (`tests/crdt-cross-wire-overwrite.test.ts`) drives the real `applySyncChange -> moveIfIdRelocated` path with a cross-wired map and asserts the unrelated note survives.

## Version

Bumps to **1.11.20**. PR #190 already claims 1.11.19 on another branch; this leapfrogs per repo convention (one bump per PR, at open time).

## Out of scope (follow-up owed)

The active-leaf editor-binding seam (`main.ts:1584`) can mint a junk `"null"` path entry when a leaf has no file yet. This PR's `NoteIdMap` garbage-path rejection (#2 above) defends against that junk key being used, but the mint call site itself is not hardened here — tracked as a separate follow-up.

## Context

Full incident forensics, prod telemetry, and root-cause writeup: `docs/context/crdt-wrong-mint-cross-file-overwrite.md` (engram-workspace repo).

## Test plan
- [x] `./node_modules/.bin/biome check src tests`
- [x] `bun run lint:obsidian`
- [x] `bun run lint:css`
- [x] `bun test` — 1668 pass, 1 skip, 0 fail
- [x] `bun run build`


Rasbandit:
Pushed `b5a8fbf` — closes both findings from the independent review of the guard commit:

1. **Stale-snapshot false-absent (data-loss path):** destructive verdicts now require a manifest snapshot ≤30s old; a failed/absent refresh drops the stale snapshot entirely (refusing is recoverable, trashing is not). Failures are negative-cached so manifest-less backends don't refetch per relocation.
2. **Refusal orphan leak:** a refused trash during a genuine rename left a duplicate no id references. Refused paths now queue for an orphan sweep re-judged against the fresh manifest on each reconcile.

Both have regression tests (`crdt-cross-wire-overwrite.test.ts`, 6 cases). Full suite 1671 green, all lints clean.

Companion backend PR (the incident trigger): engram-app/engram `fix/crdt-handshake-ratelimit` — link incoming.

Rasbandit:
Companion backend PR is up: engram-app/Engram#947 (handshake rate-limit lane — the incident trigger). Release both together.

PR #194, opened 2026-07-08
fix(crdt): editor binding must never span a file switch (round-2 cross-file pollution)

Rasbandit:
## The bug

Round-2 cross-file pollution: 2026-07-07's full content overwrote 2026-07-05 on a plain file switch — with a **proven-clean noteIdMap** (jq showed a perfect bijection), so the identity layer was exonerated. The mechanism is in the editor binding:

`EditorController.bindTo(view, newPath)` awaited `getYText(newPath)` **with the old note's ySync still attached**. That await gap is exactly where Obsidian's `loadFileInternal`/`setViewData` replaces the editor's whole document with the NEW file's content. The still-attached old ySync treats that replacement as a local edit and applies it to the **old note's Y.Text**, which then syncs up as that note's content. One file switch = one race window; content arrives clean because it's a full-doc replace.

The 3s drift-repair had the same class of hole: it dispatched repairs into the view without verifying the view still displays the bound path — a missed/late rebind meant repainting the old note's content into the visible file every 3s (the "note keeps reverting" symptom).

## The fix (Relay's semantics, ported)

Relay never lets a binding span a file load (`__relayLoading` critical section; `view.file` identity checks). Equivalent guards here:

1. **Synchronous detach**: `bindTo` clears the compartment + releases refcount *before* any await. Unbound gap = safe (worst case: no live sync until next `refresh()`); stale-bound gap = data loss. This deliberately flips the old `getYText`-rejection invariant ("old binding untouched") — that invariant WAS the bug.
2. **bindEpoch guard**: overlapping `bindTo` calls — latest path wins; a slow stale bind aborts instead of clobbering.
3. **View-identity guard on drift repair**: new `viewPath` dep (live MdView file lookup). If the view no longer shows the bound path, detach and let `refresh()` re-bind — never dispatch.

## Tests (TDD, all RED first)

- `bindTo detaches the old binding SYNCHRONOUSLY, before awaiting getYText`
- `overlapping bindTo calls: the LATEST path wins even when an earlier call resolves later`
- `drift check releases the binding when the view no longer shows the bound path`
- rejection test rewritten for the new unbound-not-stale-bound semantics

Full suite 1673 pass / 0 fail; build + biome ci + eslint + stylelint clean. v1.11.21.


Rasbandit:
Version-queue note (as of 2026-07-07 evening): main=1.11.20; **#190 and #194 both currently claim 1.11.21**; #196 claims 1.11.22. Whichever of #190/#194 merges second needs a re-bump (and #196 re-bumps if the ladder shifts). Merge order is Todd's call — this is just so nobody is surprised by the version-check.

PR #195, opened 2026-07-08
docs(context): editor-binding stale-buffer race

Rasbandit:
Repo-local context doc for the round-2 cross-file pollution discovery (PR #194): the EditorController.bindTo await gap where the old ySync binding captured Obsidian's whole-doc setViewData replace as a local edit on the OLD note, with a proven-clean noteIdMap. Records the mechanism, why the old defer-release invariant WAS the bug, the never-span-a-load fix (sync detach before await + bindEpoch latest-wins + drift view-identity guard, detach-not-release), and the fake-view testing gotcha.

- New: `docs/context/crdt-editor-bind-race-pollution.md`
- Pointer line added to CLAUDE.md `## Context Docs`
- Version bump 1.11.20 -> 1.11.21 to satisfy version-check (docs-only otherwise); whichever of this or #194 merges second rebumps per the check's race resolution.


Rasbandit:
Folded into PR #194 (one initiative, one PR) — same v1.11.21 bump would have collided at merge.

Issue #189, opened 2026-07-07
e2e live-window: test_10 fails without pull-assist — received=yes materialized=no


Rasbandit:
## Context

Part of the backend e2e testing-confidence wave (Task A1): the delivery tests were asserting via a manual `trigger_full_sync` pull before checking the receiver's vault, which masked whether live WebSocket/CRDT delivery actually worked. Converting `test_10_rename_propagation.py` to assert live delivery FIRST (no pull backstop, via `wait_for_delivery`) surfaces a real gap.

## Repro

```
cd backend/e2e
ENGRAM_API_URL=http://localhost:8100/api ENGRAM_PLUGIN_SRC=<plugin main worktree> \
  AUTH_PROVIDER=local E2E_ENABLE_CRDT=true E2E_WORKERS=1 \
  CI_POSTGRES_CONTAINER=engram-crdt-postgres-1 CI_MINIO_CONTAINER=engram-crdt-minio-1 \
  python3 -m pytest tests/test_10_rename_propagation.py -v --reruns 0
```

Reproduces on a clean `origin/main` plugin build (`77f0338`), not just a WIP branch. Confirmed with a manual 90s timeout that it never resolves (not just slow) — the new path is never written to B's vault via live delivery alone.

## Failure

```
TimeoutError: E2E/RenameNew-<suffix>.md not delivered within 30s. Client-log evidence: received=yes materialized=no.
  ws: CRDT-managed: skipping legacy body apply for E2E/RenameNew-<suffix>.md
  push: Push ok: ...
  channel: Event: upsert E2E/RenameNew-<suffix>.md
```

## Root cause

1. A renames `old_path` → `new_path`. Both devices already have this note_id enrolled in the CRDT room (B received `old_path` live before the rename, in the same channel session).
2. B's `handleStreamEvent` (src/sync.ts) runs `moveIfIdRelocated(event.id, event.path)` — this re-keys `noteIdMap` and **trashes the old local file**, but does not create a file at `new_path`.
3. Further down, since the note is CRDT-managed, the code deliberately skips the legacy body-apply (`rlog().info("ws", "CRDT-managed: skipping legacy body apply for ...")`) and instead calls `this.crdtEnrollment.enroll(noteId)`, expecting the CRDT handshake to eventually flush the body to disk via `flushFromCrdt`.
[...]

PR #196, opened 2026-07-08
fix(sync): materialize stall on received live notes (#189)

Rasbandit:
## Summary

- Root cause (confirmed by reading the code + local e2e repro, matching the issue's own analysis): a rename's WS broadcast id-relocates the note (`moveIfIdRelocated`: trashes the old path, re-keys `noteIdMap`) but carries no content change, so it never produces a Y.Doc update — `onFlushToDisk` never fires — and `crdtEnrollment.enroll()` is a documented per-session no-op once the note is already enrolled (true for a note this device saw live, at its old path, earlier this session). With neither trigger left, the new path is never written on disk: `received=yes materialized=no` (e2e `test_10`, deterministic).
- Fix: `SyncEngine.materializeRelocated()` — called right after `enroll()` in `handleStreamEvent`'s CRDT-managed branch — flushes the CRDT doc's already-known content directly to the new path when `CrdtManager.isSynced(noteId)` is true (handshake genuinely completed this session) and the file doesn't already exist. Gated on `isSynced`, not "already enrolled" (which flips true synchronously before STEP2 resolves — using it instead would race a note's genuine first enrollment and risk flushing empty/partial content, the `#547` class of bug).
- `test_43_deeply_nested_folders` (4-note burst, intermittent) is a brand-new-notes scenario, not a rename — its delivery portion passed cleanly in every local run observed (fixed or not), so it isn't exercised by this change; its remaining flakiness traces to unrelated infra (see report).

## Root cause + evidence

Full writeup (mechanism, code trace, RED/GREEN e2e logs, repro stats) at `docs/context` handoff in the workspace repo: `.superpowers/sdd/fix-189-report.md`.

**RED (before fix, plugin `main.js` built without the diff):** `test_10_rename_propagation` — `TimeoutError`, B never materializes `RenameNew-*.md` within 30s. 1/1 clean reproduction against local CRDT stack (backend `test/confidence-wave` @ 0.5.641).

**GREEN (after fix):** `test_10_rename_propagation` passed cleanly (24s, 28s) in 3/10 repeated local attempts; the other 7 were unrelated infra flakes (CDP plugin-not-ready / register connection errors from rapid back-to-back Obsidian+Xvfb spin-up on a single dev box, and one auth-token 401 at fixture setup) — **zero** of the 10 reproduced the `materialized=no` signature (grepped every log). `test_43`: 2/4 full passes, 1 unrelated infra setup error, 1 unrelated pre-existing folder-listing assertion failure that occurred *after* all 4 notes' live delivery had already succeeded.

## Note for reviewers

This branch (`test/confidence-wave`) is the plugin half of backend PR engram-app/Engram#950's e2e auto-pairing — its e2e-clerk run will pick up this fix automatically via the same-name branch pairing.

## Test plan
- [x] New unit tests pin the exact mechanism (RED before fix, GREEN after) + a guard-rail test for the `isSynced` gate
- [x] `bun test` — 1674 pass / 1 skip / 0 fail
- [x] `./node_modules/.bin/biome check src tests`, `bun run lint:obsidian`, `bun run lint:css` — clean
- [x] `bun run build` — clean
- [x] Local e2e repro against CRDT stack: RED confirmed pre-fix, GREEN post-fix (see report)

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

---

## Second fix — no echo-suppression window from no-op echo-skip pushes (Engram#944)

- Root cause (backend forensics: `docs/superpowers/sdd/task-A8-report.md` "Product-fix wave" section, workspace repo): `pushFile`'s `finally` block unconditionally called `markRecentlyPushed()` on every exit path, including the echo-skip no-op branches (local content already matches `syncState` after a pull/WS-applied write — nothing is transmitted). That opened a 5s echo-suppression window with no real push behind it. A legitimately-arriving second remote update within that window was then swallowed as `"Echo skip (recently pushed)"` — confirmed directly in the receiver's own log during the repro. Net effect: rapid double-writes from another device lose the second write (backend e2e `test_41_canvas_sync::test_canvas_modify_sync`, red on backend PR #950; server-side breadcrumbs proved both broadcasts fire, 1.1s apart).
- Fix: gate `markRecentlyPushed(file.path)` on `success` — the existing local variable that is only `true` when a push actually reached the server (REST `pushNote`/`pushAttachment` succeeded, or the CRDT path emitted). It's already `false` on the note/attachment hash-skip no-ops and on caught failures, so this is a one-line condition, not a redesign of the suppression mechanism.

### RED/GREEN
- RED: new test `echo-skip no-op push does not open the recently-pushed suppression window (Engram#944)` in `tests/sync.test.ts` — seeds `syncState` via a pull (no real push happens), calls `pushFile` (takes the echo-skip branch), asserts `isRecentlyPushed` stays `false` and a subsequent real remote update is applied not swallowed. Failed pre-fix (`isRecentlyPushed` was `true`).
- GREEN: passes post-fix.

### Verification
- `./node_modules/.bin/biome check src tests` — clean
- `bun run lint:obsidian`, `bun run lint:css` — clean
- `bun test` — 1675 pass / 1 skip / 0 fail
- `bun run build` — clean
- e2e: `tests/test_41_canvas_sync.py` (both `test_canvas_sync` and `test_canvas_modify_sync`, the exact repro) — **2 passed** against the local CRDT stack (backend 0.5.641)

Refs Engram#944.

---

## Third fix — round 2: disk-content materialize + stale-relocation guard

Backend CI evidence (Engram#950, e2e paired at plugin `a932519`) showed `test_10` still
`received=yes materialized=no` on the renamed NEW path even with the #189
`materializeRelocated` fix present, and `test_34_folder_rename_propagation::test_folder_rename_old_paths_cleaned`
regressed (`File .../Cleanup.md still exists after 30s`) — new under `a932519`, previously green.

### Mechanism (a) — test_10, isSynced-gate race

`materializeRelocated`'s `crdt.isSynced(noteId)` gate is correct (avoids the `#547`
premature-empty-write class), but on a fresh-boot receiver the rename event can arrive
before this device's own CRDT STEP2 handshake for the note_id lands *this session* —
`isSynced()` reads false, the gate declines, and nothing ever retries (no timer, no later
hook). Relying solely on the CRDT handshake to backfill the new path is unnecessary: the
OLD file's on-disk content is already real, trustworthy content (this device had it
materialized before the rename). `moveIfIdRelocated` now reads the old file's content via
`cachedRead` before trashing it, and flushes it to the new path directly via
`flushFromCrdt` — independent of CRDT session state. `materializeRelocated`'s isSynced-gated
path remains as the correct backstop for the rarer case where this device never had the
note on disk locally at all.

### Mechanism (test_34 regression) — stale/out-of-order WS event reverses a relocation

Live-repro'd locally (2026-07-08): the WS channel is explicitly unordered (documented on
`moveIfIdRelocated` itself). A duplicate/reordered upsert broadcast carrying an id's PRIOR
path can arrive *after* this device already applied a more current relocation for that id
this session. `moveIfIdRelocated`'s only precondition was "noteIdMap says this id is at a
DIFFERENT path than event.path" — true in EITHER direction — so the stale event read as a
second, backward relocation: re-keyed the map back to the old path, trashed the
just-relocated new-path file, and (via the disk-content fix above) recreated the old path
from its content. The captured docker log showed the exact signature: two distinct
note_ids fighting over the same path, and the old path recreated every ~5-10s
(`[client:vault] create path=.../Cleanup.md` repeating well after the rename).

Separately, even with `moveIfIdRelocated` correctly declining, `handleStreamEvent`'s own
CRDT branch unconditionally called `noteIdMap.set(event.path, noteId)` for path-learning —
which, since the map is a bijection, silently re-keyed the id back to the stale path anyway
and re-triggered `materializeRelocated` against it. Both call sites needed the guard.

**Fix:** a per-note_id `lastRelocationTs` map tracks the timestamp of the last relocation
`moveIfIdRelocated` actually applied; a subsequent event with an older timestamp for the
same id is ignored outright (no trash, no re-key, no materialize). `handleStreamEvent`'s
CRDT branch additionally checks that `noteIdMap.pathForId(noteId)` still agrees with
`event.path` before learning/materializing — if `moveIfIdRelocated` declined the event as
stale, the path-learning step skips too, so the canonical mapping (and file) already in
place is never undone.

### RED/GREEN (unit)

Two new TDD tests in `tests/sync-note-id.test.ts` (both RED before the fix, GREEN after):

- `moveIfIdRelocated materializes the new path from ON-DISK content directly` — asserts
  `New.md` is written from the old file's disk content even when `isSynced` reads false,
  and that the CRDT `projectedText` backstop is never called (disk-content path wins first).
  Companion test confirms the isSynced-gated backstop still fires correctly when there's no
  local file to read from.
- `moveIfIdRelocated ignores a stale/out-of-order WS event that would relocate BACKWARD` —
  drives a genuine forward relocation (timestamp=100), then a stale echo of the pre-rename
  path (timestamp=50); asserts the new-path file is never trashed and the map isn't
  re-keyed backward. Failed pre-fix (map ended up back at `Old.md`); passes post-fix.

### Verification

- `bun test` — 1678 pass / 1 skip / 0 fail (full suite)
- `./node_modules/.bin/biome check src tests`, `bun run lint:obsidian` — clean
- `bun run build` — clean
- Local e2e against the CRDT stack (`tests/test_10_rename_propagation.py` +
  `tests/test_34_folder_rename_propagation.py` together, matching the CI pairing): live-repro'd
  the `test_34` old-path resurrection once pre-fix (with disk-content materialize alone, no
  stale-event guard yet); after both fixes, 5/6 combined runs passed cleanly with zero
  recurrences of either target mechanism — the one remaining failure was the same
  pre-existing, unrelated backend-side timing flake (`Note ... still on server after 10s`,
  a server-side rename-delete-propagation delay under local dev-box load) reproduced
  identically on the unmodified baseline.

No version bump (1.11.22 stands).

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

---

## Round 3 — test_37 CRDT echo-hash hole + test_43 strand-heal one-shot drop

CI backend run 28915097812 (e2e-clerk, plugin `e7e0e4f`) surfaced two remaining defects.

### Defect 1 — test_37 content-loss (regression, attributed)

`test_append_sync` failed with `AssertionError: First append should still be present` —
the final file had the second append but not the first, a content-loss signature (not a
timeout). Empirically bisected by rebuilding the plugin at `b93d9a4` (pre-round-2 baseline),
`7d5d411`, and `e7e0e4f`, then looping `test_37` 10x solo against each on a local CRDT stack
rebuilt to match CI's exact backend headSha (`5fd5da21`, 0.5.643):

- `b93d9a4`: 0/10 failures
- `7d5d411`: 1/10 failures
- `e7e0e4f`: ~3-4/10 failures (across two 10-11 run batches)

**Root cause**: `pushFile`'s echo-hash suppression check (skip pushing when local content
already matches `syncState`'s last-synced hash) ran only in the legacy REST branch — placed
*after* the CRDT routing branch, which already returns early on success. A CRDT-managed
note's own materialize write (`flushFromCrdt` / `materializeRelocated`, e.g. discovering a
note via pull or a live rename) fires Obsidian's own `create`/`modify` vault event exactly
like a real edit. With no hash check gating the CRDT branch, that self-echo unconditionally
diffed the (unchanged) disk content into the Y.Doc via `applyLocalEdit`. Docker log forensics
on an isolated failing run showed the exact shape: B materializes a brand-new note
(`bytes=31`, base-only), immediately self-pushes it via CRDT, and — because nothing gated
that push on "has anything actually changed since last sync" — if the Y.Doc had meanwhile
advanced (e.g. a second REST-side append landing a moment later), the diff computed and
transmitted a stale/destructive delta on top of newer content.

7d5d411's own change (gating the *recently-pushed* echo-suppression window on `success`) was
narrowly correct for its target bug (Engram#944) but incidentally removed a side-effect that
had been reducing exposure to this separate, pre-existing hole — hence the jump from 0/10 to
a real, measurable flake rate rather than 7d5d411 introducing new bad behavior of its own.

**Fix**: hoisted the existing echo-hash check in `pushFile` (`src/sync.ts`) above the CRDT
routing branch so it covers both paths uniformly — a real edit still changes the content hash
and routes through CRDT normally; a self-echo of already-synced content is skipped before
ever reaching `applyLocalEdit`.

**Tests**: two new cases in `tests/sync-note-id.test.ts` (RED before the fix, GREEN after) —
one pinning that a self-materialize echo does not call `applyLocalEdit`, one guard-railing
that a genuine edit after the same discovery still does.

**e2e validation**: 15/15 clean `test_37` runs post-fix (vs the ~30-40% pre-fix rate on
`e7e0e4f`), against the same pinned backend commit.

### Defect 2 — test_43 brand-new-note burst stall (pre-existing, all 3 commits)

`test_deep_folder_multiple_notes` failed with `E2E/Depth43/Sub/B.md not delivered within 60s,
received=yes materialized=no` — one note out of a 4-note rapid-fire burst never materialized
on B, purely live (no pull backstop in this test). Traced the inbound path for an unknown
note_id: `handleStreamEvent`/`onCrdtDocReady` → `crdtEnrollment.enroll` → STEP1/STEP2 →
`onFlushToDisk` (`main.ts`) → resolves `noteIdMap.pathForId(noteId)`.

**Root cause**: when `onFlushToDisk` can't resolve a path for an id (a `crdt_doc_ready`
announce or STEP2 can race ahead of the `note_changed` broadcast that carries the id→path
association — more likely under a 4-note burst than a single note), it calls
`healUnknownNoteId`, which debounces 750ms then does **one** manifest reconcile
(`drainStrandedFlushes`). Any id still unresolved *after that single fetch* was permanently
dropped — logged once ("still no path for note_id=X after heal — retained in Y.Doc") and
never retried this session, even though the content was safely sitting in the Y.Doc the whole
time. A manifest fetch racing a beat ahead of the server durably reflecting a just-created
note (plausible under a burst, where 4 rows commit in quick succession) turns into a
permanent stall for whichever note loses that race — present since the strand-heal mechanism
shipped (plugin #187) and reproducible on all 3 bisected commits, not a new regression.

**Fix**: extracted the retry/give-up decision into a pure, standalone
`partitionStrandedFlushes` (`src/main.ts`) and changed `drainStrandedFlushes` to re-queue any
still-unresolved id for another debounced heal cycle (capped at 5 attempts, ~3.75s worst case
cumulative) instead of dropping it after one miss. A genuinely orphaned id still gives up
after the cap.

**Tests**: `tests/main-strand-heal.test.ts` (new) unit-tests the pure partition function
directly — resolved ids flush and clear their attempt count, unresolved ids retry under the
cap, an id at the cap gives up, and a mixed burst resolves each id independently (RED before
the fix — the export didn't exist — GREEN after).

**e2e validation**: local repro attempts (solo x5, plus contention runs alongside
`test_10`/`test_34`/`test_41` with 2 workers x3) did not reproduce the exact pre-fix stall —
consistent with CI's own description of it as intermittent (observed 2x). The fix is backed
by direct code-path analysis (the one-shot give-up is unambiguous in the pre-fix source) and
by 10/10 clean post-fix `test_43` runs (solo + contention combined); it was not verified
against a local repro of the specific failure the way Defect 1 was.

### Verification

- `bun test` — 1684 pass / 1 skip / 0 fail (full suite)
- `./node_modules/.bin/biome check src tests`, `bun run lint:obsidian`, `bun run lint:css` — clean
- `bun run build` — clean

No version bump (1.11.22 stands).

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


Rasbandit:
**Round 4** (artifact forensics, backend run 28915097812): test_10's stall was neither the round-2 guards nor the round-3 echo-hash gate — it was an **escaped rejection in `moveIfIdRelocated`'s file ops racing the rename's own delete tombstone**.

Trace (B, device c151add9): WS `delete RenameOld` deferred to pull → WS `upsert RenameNew` entered `moveIfIdRelocated` and suspended on the manifest await → the pull's tombstone trashed the old file mid-suspension → `cachedRead`/`trashFile` rejected → the rejection escaped (the hoisted call at `handleStreamEvent` is outside its try/catch) → the whole upsert died silently: no relocation flush, no CRDT-branch materialize, no `materializeRelocated`. The pull channel independently missed the new-path upsert (cursor stopped at the tombstone's seq), so nothing retried until the next test's FullSync 36s later.

Fix (`43074ab`): the file-op block now degrades — failed trash (tombstone won) still flushes the read content; failed read falls through to the isSynced-gated backstop. One guard in the shared function covers both callers. Two new unit tests reproduce the exact artifact sequence, red before / green after.

Rasbandit:
## Round 5 (run 28919928915) — two plugin fixes, artifact-first

**Defect A — test_34 folder-rename stall (`3975753`, PRE-EXISTING race).** Not the rename at all: the failure was the PRE-rename delivery of Note2. At 05:33:10.295 B (device `67825a01`) received both upserts in a burst; both discovery-creates raced `ensureFolder`'s check-then-create on the new `E2E/RenameFolder34` folder. The loser's `vault.createFolder` rejected `Folder already exists.`, `flushFromCrdt` caught-and-dropped the note (`received=yes materialized=no`), and only the next pull (05:33:43, ~3s past the 30s window) delivered it. All three round-brief suspects (relocation guard, degrade path, pair-crossing) acquitted — the smoking gun was the artifact line `crdt: flushFromCrdt: write failed ... Folder already exists.`. Fix: `ensureFolder` treats losing the race as success (cache re-check + already-exists match; real failures still throw), and `createFileWithFolders` degrades `File already exists.` to modify-with-same-content (same class, seen on `ChannelCatchUp2/3` + `Bulk/*` in the same run). TDD: folder-race + file-race tests RED pre-fix.

**Defect B — test_48 OAuth reconnect ×2 (`aa2f5c0`, PRE-EXISTING, #169 lineage — aeafcc7d exonerated).** Client log, device `35c1e961`: the swap's `saveSettings()` ran `setupNoteStream()` with NEW settings but the OLD `ApiKeyAuth` provider → `getMe()` resolved as the old user → `NoteChannel ctor userId=<apikey-user> vaultId=<new-oauth-vault>` → server `REFUSED JOIN sync:...` (`vault_not_found`), no client retry. The swap's own second `setupNoteStream()` then SHORT-CIRCUITED: connection key (settings-only) matched and `noteStream` was already assigned — the doomed channel survived as the live stream, `wait_for_stream` hung the full 60s, twice. Fix: the #169 short-circuit now also requires `liveConnected` (sync: topic actually joined) — a present-but-never-connected stream is rebuilt, re-resolving identity with the current provider. TDD on extracted `shouldReuseLiveStream()`.

Also joined a round-4 concatenated template literal biome flagged (pre-existing lint red on HEAD).

Gauntlet: `bun test` 1693 pass / 1 skip / 0 fail; biome + lint:obsidian + lint:css clean; build clean. No version bump (1.11.22 stands).

Rasbandit:
## Final-review fix wave (round 6)

TDD'd each finding from the final review — failing test first, then the fix, full suite green after each. All 8 findings addressed; none needed to be skipped.

- **CRITICAL-1** (relocation flush must be create-only) → `265315e` — `moveIfIdRelocated` now re-checks `newPath` for existence right before calling `flushFromCrdt`, skipping the flush if a concurrent doc-triggered write already landed there during the cachedRead/trashFile suspend window (mirrors `materializeRelocated`'s own exists check).
- **IMPORTANT-4** (revert-to-baseline swallowed on CRDT push) → `9d12d21` — a successful `consumed` CRDT push now records `fnv1a(content)` into `syncState`, so the echo-hash gate tracks last-*transmitted* content, not just last-flushed. Updated one pre-existing routing test whose two probes reused identical mocked content (now correctly echo-skipped by the fix) to use distinct content per push.
- **IMPORTANT-2** + **finding 5** (pull path must record relocation timestamps, units/ties) → `f44116f` — `applySyncChange` now passes `Date.parse(c.updated_at)` as `eventTs` to `moveIfIdRelocated`, normalized to the same epoch-ms basis the WS path already uses (`event.timestamp`, `Date.now()` at receipt). Unparseable values degrade to `undefined`, not `NaN` (which would permanently poison the guard for that id). Since `updated_at` is only seconds-precision on the wire, also widened the staleness comparison from strict `<` to `<=` so a same-second tie can't be misread as newer.
- **IMPORTANT-3** (reuse gate: everConnected not currently-connected) → `a99b86f` — added a sticky `everConnected` flag, set on first `onStatusChange(true)` per stream instance, reset only when a genuinely new stream is being built (setupNoteStream rebuild / full auth clear). `shouldReuseLiveStream` now gates on that instead of the raw `liveConnected` flag, so a transient disconnect no longer tears down a healthy CRDT stack via `saveSettings()`. Pure-function tests updated for naming; the doomed-channel case stays covered (`everConnected=false` for a channel that never joined).
- **MINOR-6** (guard-state lifecycle) → `28362c3` — `lastRelocationTs` cleared in `resetForVaultChange` (note_ids aren't unique across vaults); `strandHealAttempts` cleared at the vault-switch call site and on every reconnect (a reconnect just re-reconciled the map, so stranded ids deserve fresh retries).
- **MINOR-7** (remaining stall face) → `e49d4c9` — the cachedRead-failed catch in `moveIfIdRelocated` now kicks `void this.pull()`, so if the isSynced-gated backstop also declines, the invisible-note window is one pull instead of up to a 5-minute poll.
- **MINOR-8** (cosmetic) → `63035cc` — dropped the dead per-path ternary in the stale-echo test's cachedRead mock.

Gauntlet: `biome check`, `lint:obsidian`, `lint:css`, `bun test` (1700 pass / 1 skip / 0 fail across 90 files), `bun run build` — all green. No version bump (1.11.22 stands per instructions). Pushed to `test/confidence-wave` (now at `a7b88c1`).

Rasbandit:
Clock-base mismatch fixed in 75a4572: WS relocation staleness guard now derives eventTs from `event.updated_at` (server clock, confirmed present on both `note_changed` and `notes.batch` payloads — backend `notes.ex:broadcast_change/6`) instead of `event.timestamp` (client-receipt `Date.now()`, channel.ts), matching the pull path's basis. Also fixed the stale comment and the `<`/`<=` log mismatch. New ping-pong test RED-then-GREEN; full suite green.

PR #204, opened 2026-07-08
fix(sync): note-id map hygiene — vault wipe, rename evict, intrinsic gate

### LINKED ISSUE #200: noteIds map is not vault-scoped (main.ts loads it unscoped)
`main.ts` loads `NoteIdMap.fromJSON(data?.noteIds)` without a vault guard (~line 865) while `syncState` is scoped via `syncStateVaultId` (~line 347). Switching the plugin between vaults can carry note-id mappings across vaults — a latent cross-vault identity hazard of the same family as the 2026-07-07 incidents.

Fix: mirror the syncStateVaultId pattern for noteIds (drop the map when the stored vault id differs).

Refs: handoff A3.

Rasbandit:
## What

Three note-id map hazards, all in the cross-vault-identity family (Tranche A3 of the identity-as-CRDT handoff plus the #197 retro-review finding):

1. **Vault change wipes note identity** (closes #200). `main.ts` loaded `noteIds` UNSCOPED while `syncState` was guarded by `syncStateVaultId`, so switching vaults carried another vault's note-id mappings across — CRDT frames keyed by a foreign vault's ids, the same family as the 2026-07-07 incidents. `invalidateIfVaultChanged` now also clears the `NoteIdMap` (new `clear()`, in-place so the instance shared with main/live-views stays consistent), clears confirmed ids, and persists `noteIds: {}`.
2. **Batch server-rename evicts the old path's mapping** (#197 retro-review finding). `recordBatchPushOk`'s `server_path` branch now deletes the OLD path's map entry before adopting the winning id, mirroring `pushFile`'s sibling branch. Previously a create-race + path sanitization left a dangling mint keyed to a nonexistent path.
3. **`ensureNoteIdMapped` intrinsically gate-safe.** Its reconcile can reach `sweepPendingOrphans` → `trashFile`, so the `isSyncBlocked` check now lives inside the method instead of depending on every caller.

## Testing

TDD, 3 of 4 RED first: `tests/note-id-map-hygiene.test.ts` (vault-change wipe, same-vault untouched, batch rename evict, closed-gate no-op). Full suite 1703 pass / 0 fail; biome ci, eslint, stylelint, build all clean.

v1.11.26.


Rasbandit:
Pre-merge review complete. Findings addressed in 6b07142: the explicit picker path (resetForVaultChange) now wipes the map + confirmed ids via a shared wipePerVaultState() so the two vault-change paths cannot diverge; dead noteIds:{} saveData field dropped. Suite 1706/0.

PR #205, opened 2026-07-08
fix(sync): replay CAS + conflict routing + fresh-note handshake re-fire

### LINKED ISSUE #201: Fresh-note CRDT enrollment latency (~30s) + base_hash for batch/replay pushes
Two follow-ups from the 2026-07-07/08 live testing:

1. A brand-new note sits un-enrolled for ~30s+ (confirmation + enrollment pacing) before live sync engages — target sub-5s on an idle connection (handoff A6).
2. `base_hash` CAS (v1.11.23) covers `pushFile` only. Batch falls back to per-file push on conflict (which then carries base_hash) — verify that chain with a test, or add per-entry base_hash (handoff A4).

Refs: engram-workspace docs/superpowers/plans/2026-07-08-identity-as-crdt-phases.md.

Rasbandit:
## What

A4 + A6 of the identity-as-CRDT handoff. Closes #201.

**A4 — the CAS gate now covers every REST push path:**
- **Offline-replay pushes declare `base_hash`** (stored `syncState` version + serverHash, same 3-shape call cascade as `pushFile`). Previously a replay sent neither version nor base, so a stale edit queued during an offline window sailed past the v0.5.642 CAS gate and overwrote server content this client never saw — the incident class.
- **A conflicting replay routes into the single-note conflict flow** (`pushFile(file, true)` → 3-way merge / conflict-copy). Previously a conflict response was silently dequeued: the local edit vanished from the pipeline with no handling at all.
- **The batch path is pinned by test**: a per-entry `conflict` falls back to `pushFile`, which carries `base_hash` — the chain the handoff asked to verify.

**A6 — fresh-note enrollment latency (~30s → immediate):**
A new note's pre-push STEP1 is dropped server-side (`note_not_found`, no row yet) and the once-per-session enrollment guard never re-fires it — the note stays deaf to live sync until a later catch-up. New `refireEnrollmentOnFirstConfirm`: on the FIRST confirmation of an id (the create-push response, `pushFile` both branches + replay adopt), `reset` lifts the guard and `enroll` re-fires STEP1 now that the row exists. Md + size gated exactly like the pre-push enroll; already-confirmed ids never re-fire (no handshake storms on bulk pushes).

## Testing

TDD: 4 of 7 tests RED first (`tests/sync-replay-cas-enroll.test.ts`). Full suite 1715 pass / 0 fail; biome ci, eslint, stylelint, build all exit 0.

v1.11.27.


Rasbandit:
Pre-merge review complete: safe to merge. Traced and cleared: STEP2 idempotency (merge, never replace), pull-races-push guard, arguments.length pins, handshake-storm bound (refire not wired to batch — deliberate), replay re-entrancy/double-dequeue, deleted-file handoff. One minor double-fault window (flush-loop dequeue clobbering a mid-resolution retry) filed as a follow-up issue — the naive gate would strand keep-remote resolutions in infinite retry, so it needs a return-shape change, not a one-liner.

Issue #203, opened 2026-07-08
verifyConvergenceOnOpen false-positives on CRDT-live notes (spurious WARN + redundant re-handshake)


Rasbandit:
Retro-review finding on merged PR #198 (Important, no data risk).

**Root cause:** `syncState.serverHash` is advanced only by REST push responses and the pull backfill — the live CRDT flush path never maintains it (and cannot trivially: content_hash is a server-side HMAC). For an actively-CRDT-edited note, `manifestPathHashes` holds the current server hash while `syncState.serverHash` stays frozen at the last REST/pull value, so `verifyConvergenceOnOpen` reads permanent divergence: every `file-open` of a HEALTHY note logs `bind-time divergence` at warn and fires a redundant reset+enroll re-handshake.

**Impact:** alert-fatigue on the diagnostic tripwire + wasted handshake round-trips. Data-safe (STEP1/STEP2 is idempotent). Self-heals per note only after it's closed and a later pull backfill rewrites serverHash.

**Fix directions (needs a design pass, not a hack):**
- have the CRDT flush path record convergence (e.g. onFlushToDisk learns the checkpoint hash via the announce/changes metadata), or
- skip/soften the check for notes with an active live binding or recent CRDT activity, or
- rate-limit the forced re-handshake per note per TTL window and downgrade the log to info when live-enrolled.

**Also from the same review (low-confidence race, wants an e2e case, ties into handoff A5):** the pull backfill overwrites local disk for a CLOSED note; if that note's Y.Doc carries a local edit not yet reflected in the pulled content, ordering matters — add 'local edit on closed note races divergent pull' to the e2e suite.

Refs: engram-workspace docs/superpowers/plans/2026-07-08-identity-as-crdt-phases.md.

Rasbandit:
Severity upgrade: this gap is not just spurious re-handshakes — it was a live data-loss hole. e2e test_83 (backend PR incoming) proved it against v0.5.643 + plugin v1.11.27: a device that received a note ONLY via the C1 WS-event branch had no recorded serverHash at all, so after a missed delivery its REST-fallback push declared no base_hash, passed the CAS gate, and silently erased the server edit (no conflict copy, zero note_stale_base_rejected).

PR #207 fixes the C1 leg: the branch now records serverHash + version from the WS event. With it, the same e2e run 409s the ignorant push and preserves both edits.

[...]

PR #207, opened 2026-07-08
fix(sync): C1 WS-event branch records the CAS base (missed-delivery data loss)

Rasbandit:
## What

Closes the live data-loss hole found by the new e2e incident-chain test (identity-as-CRDT handoff A5, backend PR to follow):

`handleStreamEvent`'s C1 branch (WS upsert for a CRDT-managed md note) skips the legacy body apply and delegates to CRDT delivery — but recorded **no syncState at all**, and the CRDT delivery path never advances `serverHash` (issue #203). So a device whose only knowledge of a note came through C1 had no CAS base. Its later REST-fallback push (channel down — exactly the missed-delivery scenario the CAS was built for) declared no `base_hash`, passed the v0.5.642 gate untouched, and **silently erased server content the device never saw**. e2e `test_83_missed_delivery_no_deletion` reproduced this deterministically against v0.5.643 + plugin v1.11.27: server edit gone, no conflict copy, zero `note_stale_base_rejected` server-side.

**Fix:** the WS event carries the REST-level `content_hash` + `version` — record them in the C1 branch (local `hash` preserved, or seeded empty for a never-seen note, which only biases toward pushing — the safe direction).

**Validated:** with this fix, the same e2e run logs exactly one `note_stale_base_rejected`, the auto conflict flow preserves both edits, and the device converges on reconnect. Both new e2e tests (create-race, missed-delivery) pass.

Note: the pure-CRDT-frame leg still never advances `serverHash` (a CRDT-sourced edit doesn't bump the REST hash until checkpoint) — that staleness errs toward a false conflict, not data loss. That remaining design pass stays with #203.

## Testing

TDD: 2 tests RED first (`tests/sync-c1-serverhash.test.ts`). Full suite 1719 pass / 0 fail; biome ci, eslint, stylelint, build all exit 0. Live-validated via the e2e stack (test_82 + test_83 green against this build).

v1.11.28.


Rasbandit:
Review finding addressed in the follow-up commit: the write is now SEED-ONLY, gated on `prior?.serverHash === undefined`. This protects existing converged bases (the reviewer's silent-divergence regression) while still covering the raced ordering the suggested `!getFileByPath` gate would miss — flushFromCrdt can write the file to disk (recording only a local hash) before the WS event processes, and file-existence gating would leave that entry baseless and blind. Third test added pinning exactly that race; the no-clobber test pins the converged-base case. Suite 1720/0; both e2e incident tests re-validated green against this build.

Rasbandit:
Heads-up: #209 (just merged, v1.11.31) rewrote the adjacent C1 catch-up branch in `applyChange` — the backfill now requires a provably-clean local file (`fnv1a(local) == stored.hash` or local == remote) and routes genuine divergence to the legacy conflict machinery. This PR touches the same C1 code (`fix/c1-serverhash-cas-base`) and should merge main + reconcile before landing. Context: main run 28926366421's test_14 failure → latent-since-#198 data-loss hole, forensics in engram-workspace `.superpowers/sdd/fix-189-report.md`.

PR #211, opened 2026-07-08
fix(sync): materializeRelocated identity re-check — no tombstone resurrection

### LINKED ISSUE #210: test_34: stale old-path upsert resurrects a tombstoned path via materializeRelocated (fresh-mint re-push)
CI-deterministic since reruns=0 (4/4 red on runs 28926868261/28926900381/28930769463 + reruns); passes locally with workers=1. Artifacts from run 28930769463, B-device timeline for E2E/RenameCleanup34/Cleanup.md:

1. Folder rename: the tombstone DELETE for the old path arrives BEFORE the new-path upsert → "Delete deferred to pull (live note at path)" (correct).
2. A trailing OLD-PATH upsert event then hits the C1 branch. The map still says the old path is canonical (the pull's id-keyed move hasn't run yet), so #190's stale-path guard passes it, and #196's materializeRelocated (unawaited) fires.
3. The pull's "Id-keyed move: old → new" relocates the mapping and cleans the old file — then the stale materializeRelocated write lands and RE-CREATES the old-path file (`[client:vault] create path=old`).
4. The re-created file's modify event pushes with a FRESH MINT (log shows a second id: 019f40f8-c886-... vs original 019f40f8-c658-...) because the old-path map entry moved away — resurrecting the tombstoned path server-side. Old path never cleans; wait_for_file_gone times out.

Fix shipped in the companion PR: materializeRelocated re-checks id→path identity at WRITE time (after its file-missing check): if `noteIdMap.pathForId(noteId)` no longer equals the path it captured, the id has relocated and the write is stale — skip. No new state, closes the race at the only door this trace used.

Rasbandit:
## What

Closes #210 — the last member of the reruns=0-unmasked set (e2e test_34, 4/4 CI-deterministic red).

`materializeRelocated` (#196) is unawaited and races the pull's id-keyed move. CI artifacts (run 28930769463, B-device timeline) show the full loop: the rename's tombstone arrives before the new-path upsert (deferred to pull, correct) → a trailing old-path upsert passes #190's stale-path guard (the map hasn't relocated yet) and schedules the materialize → the pull relocates the id and cleans the old file → the stale write re-creates the tombstoned path → its modify event pushes it under a **fresh mint** (a second id is visible in the logs) — resurrecting the path server-side, so `wait_for_file_gone` times out forever.

**Fix:** re-check id→path identity at write time, after the existing file-missing check: if `noteIdMap.pathForId(noteId)` no longer equals the captured path, the id relocated and the write is stale — skip (logged as `Stale materialize skipped`). No new state; mirrors the #190 guard's semantics at the other end of the race.

## Testing

TDD: RED first (`tests/sync-stale-materialize.test.ts` — relocated-id skip + still-canonical positive control). Full suite 1758/0; biome ci, eslint, build exit 0. Rename-family e2e (test_10 + both test_34 tests) green locally against this build. The CI ordering itself is only reproducible under parallel churn — this PR's own backend/e2e run is the real proof.

v1.11.33 (1.11.32 is held by #207).


Rasbandit:
Review finding applied: the identity re-check now runs AFTER the projectedText await, immediately before flushFromCrdt — the relocation can land during that suspend and flushFromCrdt fails open, so the pre-await placement left the exact window the fix targets. Also documented that the guard defends the MOVE case only (tombstone delete clears byId → canonical null; the isSynced teardown gate is that backstop). Suite 1762/0.

Issue #159, opened 2026-07-02
crdt: pace pull-discovery enrollment for large fresh installs (240/10s crdt_msg limit)


Rasbandit:
From PR #155 final review (E). A fresh install of a big vault mass-enrolls every discovered note during the first pull — 2+ frames per handshake trips the backend's 240/10s per-user crdt_msg limit (engram#846) at ~120 notes. With #155's ref-matched join errors this degrades gracefully (dropped STEP1s self-heal via later fullSync), but initial sync latency on 1000+-note vaults would benefit from a paced enroll queue (~20/s).

PR #213, opened 2026-07-08
fix(sync): back off crdt: join retries; audit stale-token WS reauth loop

Rasbandit:
## Summary

Two plugin-side fixes for engram prod alert-noise root causes (see workspace `alert-noise-investigation.md`, loki-error-rate + auth-failure-burst sections).

### Fix 1 — CRDT channel-join retry loop needs backoff

A `crdt:` topic join rejected mid-session (e.g. `"rate_limited"` from the backend's join/crdt_msg limiter — #159 context: 240/10s bucket) doesn't force a reconnect by itself, but when the whole socket then drops, the existing post-open reconnect path (`channel.ts` `onclose`) used a **flat full-jitter window that never grows across repeated failures** — unlike the pre-open path, which already exponentially backs off 1s→60s. A server that keeps closing the socket right after rejecting the join produced a tight open/reject/close loop: exactly the prod burst pattern (237/131/123 client-category error lines).

Fix: track whether the `crdt:` join was rejected in the session that just ended (`crdtJoinFailedReason`). If so, the next reconnect uses a growing backoff (`joinFailureBackoffMs`, capped at 60s) instead of the flat jitter, with a higher floor (10s, `RATE_LIMITED_JOIN_FLOOR_MS`) for an explicit `"rate_limited"` reason. A graceful drop with no join failure is unaffected (existing behavior preserved). Reset to the floor on the next successful `crdt:` join.

### Fix 2 — stale-token WS reauth loop (investigated, no code hole found)

Traced the full reconnect-after-auth-reject path: `channel.ts` `onclose` fires an authenticated `GET /me` probe on a fast pre-open close (`setAuthProbe`, wired at channel construction before `connect()` — no boot-time gap); `api.ts`'s real-401 handler is the sole invalidation authority, invalidating + retrying once with a freshly-refreshed OAuth token; repeated failures back off exponentially (1s→60s) or defer 30s (`NO_AUTH_RECONNECT_MS`) once the refresh token is confirmed dead. This chain is the deliberate result of #129 (backend-switch in-memory provider leak) + #178 (network-blip vs auth-reject disambiguation, merged 2026-07-06) — the fire-and-forget probe racing the reconnect timer is an intentional, documented tradeoff from #178 (awaiting it risks the zombie-socket class the single-flight guard exists for), not a bug.

Queried prod Loki directly: the `signature_error` bursts carry `alg=RS256, iss=https://clerk.engram.page` (Clerk-templated OAuth tokens, confirming SaaS users, not self-host). Client-log metadata never carries a plugin version (only Obsidian's own `user-agent`, e.g. `obsidian/1.12.7`) — so per-burst plugin-version attribution isn't directly provable from Grafana today. What IS provable: the Jul 5 23:11 burst (43 lines) predates #178's merge (Jul 6 01:08), confirming at least that portion is old-version residue. No remaining current-code hole reproduces an infinite identical-token loop, so no plugin change is made for Fix 2 — YAGNI per the ticket's own escape hatch.

## Test plan
- [x] TDD: 5 new failing→passing tests in `tests/channel-join-backoff.test.ts` covering growth, cap, rate_limited floor, and reset-on-success; regression test confirms the graceful-drop path is untouched
- [x] `bun test` — 1774 pass, 1 skip, 0 fail
- [x] `./node_modules/.bin/biome check`, `bun run lint:obsidian`, `bun run lint:css` — clean
- [x] `bun run build` — clean
- [x] Version bumped 1.11.34 → 1.12.1 (leapfrogs open PR #192's stale 1.12.0 claim off an older base)


Rasbandit:
## Attribution: red e2e checks are pre-existing flakes, NOT caused by this PR

Downloaded artifacts + full job logs for the triggered run (`engram-app/Engram` run [28979516053](https://github.com/engram-app/Engram/actions/runs/28979516053)) and checked the reviewer's join-rejection-backoff hypothesis directly against the logs.

**Direct evidence against the hypothesis:** grepped both jobs' `docker-compose.log` (which re-emits every remote-logged plugin message, including this PR's new `crdt: join rejected (...) — backing off reconnect ...` line) for `join rejected` / `backing off reconnect`. **Zero occurrences in either job.** The new backoff branch (`channel.ts` `if (opened && this.crdtJoinFailedReason !== null)`) never fired in this run — it cannot be the cause of any of the four failures.

Per-test attribution:

- **`test_concurrent_edits_both_survive`** (e2e-crdt) — failed in **~1.8s** (test 1 finished 22:16:40.45, this one's FAILED marker at 22:16:42.21). Far too fast to involve any 10-60s backoff. `a_final == 'shared base\nFROM_B\n'` — A's own local edit got clobbered, a genuine CRDT merge/race bug, unrelated to reconnect timing. Zero join-rejections logged anywhere in the e2e-crdt job.
- **`test_49::test_rapid_api_edits_content_not_stale`** — asserts `cdp_b.check_stream_connected()` is true *before* the loop (passed), then drives 7 rapid **REST** `POST /api/notes` calls and waits for the content to land — no CRDT channel join involved. This is the same signature as closed issue [#869](https://github.com/engram-app/Engram/issues/869) ("Clerk POST /v1/sessions 404 exhausts retries in test_49 cross-auth tests") — a known recurring test_49 flake class.
- **`test_84`/`test_85`** — both are the **known pre-existing "last-in-suite churn" flake class**, already being actively mitigated on `main` independent of this PR: [#963](https://github.com/engram-app/Engram/commit/) "test(e2e): force-connect precondition for test_84/85 (last-in-suite churn)" (v0.5.647) — its own commit message: *"these tests run last and inherit the whole suite's socket churn, so the stream can sit mid-backoff at test start."* The force-connect precondition from that fix is present in the test code that ran here (confirmed in the job log) and still didn't fully close the flake — this run is another instance of that same pre-existing class, not a regression from this PR.

**Conclusion:** all four failures are unrelated pre-existing flake classes. No design change made to `channel.ts` — the backoff logic (10s floor only on `rate_limited`, exponential otherwise, reset on success) is unchanged from the last review pass, since there is no evidence it caused any observed failure.

Reran the failed jobs: `gh run rerun 28979516053 --failed`.

PR #215, opened 2026-07-08
fix: suppress self-echo of wipeRemote deletes (replace-remote vault wipe)

Rasbandit:
## What happened

`Delete all on remote, then upload local files` (the sync-preview modal's replace-remote choice) deleted **everything on both sides** (2026-07-08 incident, selfhost dev):

1. `wipeRemote()` REST-deletes every remote note while the WS channel stays connected.
2. The server fans each delete back to the deleting device (`note_changed op=delete mode=fanout` — broadcasts carried no origin attribution).
3. The stream handler deliberately exempts deletes from push-echo suppression (renamed-away-path rationale, e2e test_10), and the confirmed-note guard doesn't hold on a fresh device — so every echo hit `trashFile()` and emptied the local vault mid-wipe.
4. `pushAll` then enumerated an empty vault: `pushed=0`, both sides gone.

## Fix (two layers)

**Layer 1 — wipe path marking.** `wipeRemote()` marks each path (`markWipedRemote`, 60s TTL) **before** issuing the REST delete; the stream delete branch skips marked paths. Also clears the wiped path's server bindings (syncState, baseStore, noteIdMap, CRDT doc + enrollment) so the follow-up push mints every file as **new** — a retained serverHash would hash-skip "unchanged" notes into silent remote loss, and retained note_ids join-spam dead CRDT rooms. Works against ANY backend version.

**Layer 2 — origin-attributed echo drop (#970 client half, paired with engram-app/Engram#969).** The server now stamps the REST caller's `X-Device-Id` (which the plugin already sends) into delete broadcasts; the engine drops delete events carrying its own device id. Generic class fix for REST-driven self-echo — kept **in front of** layer 1, which remains for pre-#970 self-host backends. Upserts intentionally untouched: their echoes drive id-relocation and already have hash/pushing suppression.

Scoped: deletes for non-wiped, foreign-device paths still apply (cross-device delete propagation and rename cleanup unaffected — e2e test_05 + test_10 green).

## Tests

- `tests/sync-wipe-remote-echo.test.ts` — 6 unit tests: echo skip for notes + attachments, scoping, binding teardown, own-device drop, foreign-device apply. Red before each layer, green after.
- Paired e2e (engram branch `fix/replace-remote-self-echo-wipe`): `test_86_advanced_sync_choices.py` covers all four advanced-sync buttons end-to-end; the replace-remote test reproduced the incident on 1.11.34 and passes with this fix. Full plugin suite 1776 green; eslint/stylelint/biome ci clean. v1.11.35.


Rasbandit:
Multi-agent review round (high effort, 18 verified findings → 10 distinct) applied in `00ea482`: unconditional binding teardown in wipeRemote (timeout-masked deletes no longer cause both-sides loss), foreign-device-attributed deletes bypass the wipe guard, editor bindings detach before Y.Doc teardown (new `CrdtLiveViews.detachAll`), noteIdMap clear covers .canvas, TTL 60s→10min for pre-#970 backends, `markWithTtl` dedupe + `destroy()` timer sweep. All TDD'd — suite now 1780 green.

PR #220, opened 2026-07-09
fix: materialize web-created empty folders live via folders.batch

Rasbandit:
## Problem

An empty folder created in the web app never appeared in Obsidian until the next pull. Two stacked gaps:

1. Backend `POST /api/folders` never broadcast (fixed in engram-app/engram — companion PR on branch `fix/web-display-filename-not-title`).
2. The plugin had **no handler** for `folders.batch` at all, so even the existing batch delete/move broadcasts went unheard.

## Fix

- New `onFoldersChanged` callback on `NoteChannel`, dispatched for any `folders.batch` event (create/delete/move).
- Wired in `main.ts` to `SyncEngine.resyncFolders()` — a thin public wrapper over the existing idempotent `syncExplicitFolders()` poll (re-fetch `/folders/explicit`, materialize on disk, hydrate the explicit-folders set).

## Not in scope

Web-deleted empty folders are still not removed from Obsidian live — the resync only creates, never deletes (auto-deleting a local folder that may hold unsynced files needs its own design).

## Tests

- Channel dispatch tests: `folders.batch` create/delete/move fire `onFoldersChanged`; `onEvent` untouched.
- Full suite green (1800 tests), build + eslint + stylelint + biome clean.

Requires the backend companion PR for live *create* events; delete/move broadcasts already exist server-side.


Rasbandit:
Added a second fix to this PR (same live-sync initiative): **connectChannel's getMe() preflight gave up permanently after 5 attempts (~30s)** — any backend outage longer than that killed live sync until plugin reload, while REST recovered on its own. Forensics from the selfhost dev backend's relayed client logs: `connectChannel(attempt=4)` → `getMe() failed (attempt 5/5)` → silence, then hours of successful token refreshes and pulls with zero socket connects. Retry is now indefinite with the same capped-60s exponential backoff NoteChannel uses post-drop, still epoch-guarded.

Issue #221, opened 2026-07-09
[p0] CRDT catch-up convergence + lazy room enrollment


Rasbandit:
## Incident

Syncing phone + laptop on a ~230-note vault put a user's vault into a non-converging CRDT catch-up loop (2026-07-09 ~07:26–07:53 UTC, plugin `1.12.6`). An open (live-bound) diverged note re-handshaked every 5 minutes forever, and each reconnect re-enrolled the entire vault, driving a backend DB-pool meltdown (see engram backend issue). Affected notes: `Journal/Labs & Tests/*`, `Journal/Health Journey.md`, `Knowledge Vault/Meeting Notes/EoE…`. No data lost.

## Root cause (B2): live-bound catch-up never records convergence

- Divergence is a server `content_hash` vs stored `serverHash` mismatch in `applyChange` (`src/sync.ts:3519-3520`).
- Two branches. The **non-live-bound** backfill branch converges correctly: it writes `serverHash: change.content_hash` (`src/sync.ts:3564-3568`).
- The **live-bound** branch (`src/sync.ts:3521-3533`) calls `crdtEnrollment.reset(noteId)` + `enroll(noteId)` to re-handshake but **never writes `serverHash`**. The follow-up flush `recordCrdtBaseline` (`src/sync.ts:792-795`) sets `hash` but not `serverHash`.
- So `stored.serverHash` stays stale permanently. The 5-min background poll (`src/main.ts:1983-2006`, `FALLBACK_POLL_MS`) re-runs the same check, re-fires `forcing re-handshake`, forever. No convergence check, no backoff, no give-up.

## Secondary: eager full-vault enrollment (scaling cliff)

- On connect the plugin fires one `STEP1` handshake per note in the vault (mass enrollment), and the changes feed is applied in a **delay-free serial loop** (`src/sync.ts:3360-3373`) with **zero client-side throttle** on `sendCrdt` (`src/channel.ts:255-263`). ~20 handshakes in ~3s observed.
- This makes cost track **vault size** instead of **active editing**, and turns every reconnect into a whole-vault room storm on the backend. This is the structural scaling problem (5 users × 1k notes).

## Tasks

- [ ] Fix convergence: advance `serverHash = change.content_hash` after a live-bound re-handshake completes (mirror the non-live-bound branch). Smallest correct fix for the incident loop.
[...]

PR #222, opened 2026-07-09
fix(crdt): converge catch-up loop + lazy enrollment (flag-off)

Rasbandit:
## What

Two things: (1) fixes the non-converging live-bound CRDT catch-up loop (the engine of the 2026-07-09 sync incident), and (2) adds lazy room enrollment behind a default-off flag (the scaling fix). Backend half: engram-app/Engram#983.

## 1. Converge live-bound catch-up (default ON, the loop-killer)

Divergence is a server `content_hash` vs stored `serverHash` mismatch. The non-live-bound branch converges (writes `serverHash`); the **live-bound** branch re-handshaked but never advanced `serverHash`, so every 5-minute poll re-fired the same handshake forever. That perpetual re-handshake drove the reconnect storm that exhausted the backend DB pool. Fix: record the reacted-to server hash after a live-bound re-handshake, so the next poll recognizes convergence. Local `hash` is preserved (the open editor owns the body; no disk write).

> Note on scope: I intentionally did **not** add a re-handshake backoff/cap. With convergence recorded, a live-bound note re-handshakes only when the server hash actually changes (legitimate), and `verifyConvergenceOnOpen` is already the backstop for any missed delivery. Capping would suppress correct behavior. The whole-vault handshake burst is the *eager enrollment* problem, addressed below.

## 2. Lazy room enrollment (default OFF, the scaling fix)

Adds `lazyEnrollment` (default `false`). When on, the plugin only opens a live CRDT room for a note open in the editor; cold notes sync via the room-free REST pull feed (receive) and convergent REST push (send). Sync cost then tracks active editing, not vault size, so a large vault no longer opens hundreds of rooms on connect (the storm amplifier).

Seams gated when lazy:
- the two eager pull-loop `enroll` sites,
- the cold-start reconcile block (it seeds drift into the Y.Doc for every drifted file via `applyLocalEdit`, bypassing the push gate),
- **the correctness hinge:** both CRDT push gates additionally require `isLiveBound`. The gate keys on `isNoteConfirmed`, and every pulled note is confirmed, so without this a confirmed-but-cold edited note would seed a **duplicate** Y.Doc lineage (#846/#161 class). Under lazy it routes to convergent REST instead.

**Ships default-off** pending review and a staged rollout: sync-engine behavior changes are exactly what has caused past regressions, and this deserves a deliberate enablement rather than an auto-flip. Merging does not touch prod (releases are tag-gated).

## Tests

- Loop fix: a live-bound diverged note re-handshakes ONCE across two identical-hash polls (was: every poll forever).
- Lazy: a cold diverged note backfills via REST but is NOT enrolled; a confirmed-but-not-live-bound edit routes to REST, not CRDT.

Full suite `1799 pass, 0 fail`; `biome ci`, `eslint` clean; build (tsc) green. Version 1.12.5.

Refs #221 (loop fix complete; lazy enrollment lands flag-off, rollout pending).


Rasbandit:
Addressed the scale/reliability code review:
- **Bounded re-handshake retry** — the convergence fix recorded serverHash after one fire-and-forget handshake, so a dropped STEP1/STEP2 was never retried (open editor could miss the other device's edit). Now retries up to a cap before recording convergence: bounded (no storm) yet delivery-retrying.
- **Real local hash** on give-up instead of a `0` sentinel that a later cold-note check misread as local-diverged → spurious conflict file.
Full suite green (1803).

PR #223, opened 2026-07-09
feat(crdt): make lazy room enrollment unconditional

Rasbandit:
Makes lazy CRDT room enrollment the **only** behavior — removes the `lazyEnrollment` settings flag entirely. Follow-up to #222 (which shipped lazy behind a default-off flag). Sole prod user, hard commit.

## What changed

A live CRDT room is opened **only** for editor-open (live-bound) notes. Cold notes sync via the room-free REST pull feed (receive) and convergent REST push (send), so sync cost tracks active editing, not vault size — closing the 2026-07-09 reconnect-storm amplifier for good.

The four flag seams collapse to an unconditional `isLiveBound` check:
- **pushFile CRDT gate** — a confirmed-but-cold edit routes REST, never seeding a duplicate Y.Doc lineage (#846/#161 class). This is the correctness hinge.
- **pushNotesViaBatch CRDT-owned skip** — cold notes reach REST (nothing else would deliver them).
- **two pull-discovery enroll seams** — no STEP1 per note on connect.

The `main.ts` cold-start reconcile block is **removed**: it seeded on-disk drift for *every* markdown file into the Y.Doc, bypassing the live-bound gate. Cold-note drift is reconciled by the regular REST fullSync instead.

`lazyEnrollment` is dropped from `EngramSyncSettings` + `DEFAULT_SETTINGS`. Stored `data.json` values are harmlessly ignored (never persisted true — was default-off).

## Tests

Unit suite reworked for always-lazy semantics: `handleModify` of a live-bound note is editor-owned (early return), so the CRDT push branch is now exercised via `pushFile` directly; cold notes assert REST + no-enroll. **1801 unit tests pass**, build + biome ci + eslint + stylelint clean.

**Local e2e (CRDT stack, plugin worktree):** 7 tests green — append sync, SSE live A→B, modify/delete/rename propagation, rapid edits, conflict skip. These are all cold-path (notes never opened in the editor), directly validating the cold→REST behavior lazy now governs.

Refs #221


Rasbandit:
Closing without merge. Root-cause review found this whole lazy-enrollment direction fights a deeper problem: Engram runs TWO note-content convergence authorities (the CRDT Y.Doc AND the REST serverHash/base_hash CAS), and every bug here lives at their seam. A base_hash CAS is incompatible with CRDT by construction. Superseded by a Relay-modeled sync-core redesign (single Y.Doc authority + vault head-index scale layer); design spec forthcoming. The acute pool incident is already handled by #984/#222, so nothing is regressed by closing this.

PR #230, opened 2026-07-11
feat(crdt): make lazy enrollment the only path (remove flag)

### LINKED ISSUE #221: [p0] CRDT catch-up convergence + lazy room enrollment

Rasbandit:
## Summary

- Removes the dark-launched `lazyEnrollment` flag and hard-wires lazy CRDT enrollment as the only path. **This is the actual fix for the connect storm**: lazy enrollment shipped default-off in #222 and was never flipped, so every client ran the eager path that opens a CRDT room per note on connect.
- Cold notes never open a room on connect (they materialize room-free via the REST `/sync/changes` feed). The push gate `isCrdtManagedOffline` now requires `isLiveBound`, so a disk edit to a **not-open** confirmed note routes convergent REST (base_hash CAS), not CRDT ops, which prevents seeding a duplicate Y.Doc lineage (#846/#161 hinge). Open-note edits are unchanged (the CodeMirror binding streams them live).
- Deletes the eager cold-start reconcile block + the now-dead `reconcileColdStart` helper (closed-note drift reconciles via REST `fullSync`), and migrates the eager-era test suite to the lazy model. Net **+134 / −521**.

## Test plan

- [x] `bun test` — **1881 pass, 1 skip, 0 fail**
- [x] `tsc` + esbuild build, `biome ci`, `lint:obsidian` (eslint), `lint:css` (stylelint) — all green
- [x] Two independent reviews (correctness + test-coverage). Fixes applied: stale log line; two "offline capture" tests that were passing **vacuously** (live-bound + `handleModify` early-returns) switched to `pushFile`; added a guard test for the `handleModify` live-bound early-return (the anti-churn gate) and **verified it has teeth** — disabling the gate fails exactly that one test.
- [x] Both sides of the routing gate pinned: confirmed+live-bound → CRDT `applyLocalEdit`; confirmed+not-live-bound → REST `pushNote`. Cold-note `enroll` NOT called asserted on both pull paths.
- [ ] **Manual smoke (reviewer):** load the build, connect a large vault, confirm process count rises only by the number of editor-open notes and there are **no** `note_not_found` cold-note `crdt_msg` drops on connect.

**Behavior-change note:** a disk edit to a NOT-open note now routes legacy REST instead of CRDT ops. An external write that *preserves mtime* (`rsync --archive`, `touch -d`, backup restore) to a closed note is skipped by the mtime-gated REST reconcile that replaces `reconcileColdStart`'s hash compare — pre-existing behavior for non-CRDT vaults and the flag-on lazy path, not introduced here.

Closes #221.


Rasbandit:
Superseded by the bounded-enrollment approach (keeps all notes CRDT, throttles the connect fan-out instead of routing cold notes to REST). The lazy-only approach here broke cold-note merge in e2e (test_concurrent_edits_both_survive). See the new PR.

Issue #191, opened 2026-07-07
channel: reset crdtJoined on unclean close before sync-topic join


Rasbandit:
## Edge case

If the `crdt:` join ack lands but the sync-topic ack never does (`connected` still false) and the socket then drops uncleanly, `onclose`'s `setConnected(false)` is a transition-gated no-op, so `crdtJoined` stays `true` across the reconnect.

On the new socket, `sendCrdt` then permits pre-ack sends (it reads `crdtJoined`), and the join-ack handler's `!this.crdtJoined` guard (`src/channel.ts` ~:558) skips re-firing `onCrdtJoined` for the new session because the flag was never reset.

## Context

This is a pre-existing edge, not introduced by #190. It predates #190's `sendCrdt` gate, which now reads the `crdtJoined` flag and is therefore affected by it going stale.

## Suggested fix

Also reset `crdtJoined = false` in `onclose`, alongside `setConnected(false)`, so a reconnect always re-establishes the join-ack contract before `sendCrdt` permits sends.

Found during #190 final review.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #242, opened 2026-07-14
fix(crdt): live-bound catch-up converges via REST, never fakes convergence

Rasbandit:
## The bug (2026-07-14 live-sync deaf-note incident)

A **diverged live-bound note** (open in the editor) had two broken recovery legs:

1. `applyPushedNoteUpdate` discards vault-channel fan-out frames for live-bound notes **by design** (the room owns them) — silently, with no trace.
2. The pull catch-up path re-handshook up to 3 times, then **recorded convergence without the data ever arriving** (the bounded give-up that stopped the 2026-07-09 re-handshake storm). If the room broadcast was lost (stale observer registration after a socket drop / prod node roll), the note went **permanently deaf to remote edits** until an Obsidian restart.

Verified live in prod: web-originated edit applied server-side + fanned out (web received its own echo in 130ms), plugin connected and logging `CRDT catch-up: diverged + live-bound, re-handshake 1/3` every 5-minute poll, file never updated.

## The fix

The live-bound catch-up now **converges deterministically over REST**: pull the delta since our real state vector (`GET /notes/:id/updates`), apply it to the live Y.Doc (the editor binding paints it — no disk write), verify no pending gap, and only then record convergence. The re-handshake is kept to re-register the room observer for future live updates. A REST failure retries at the poll cadence (bounded, not a storm) and **never fakes convergence** — the give-up cap is gone because the loop now terminates by actually delivering the ops.

## Tracing (the silent paths, made loud)

- fan-out skip for live-bound notes logs at info (`client_logs`, queryable per device)
- `sendCrdt` refusals (crdt topic not joined) warn instead of silently dropping STEP1/updates
- push beacons carry `engram.note_id` + `engram.route` (UUIDs collapsed to `:id`; server-side sanitizer allowlist lands in the paired backend PR)

Pairs with backend PR (crdt_msg ack + browser telemetry + Tempo noise cut). The plugin tolerates the new ack: an ok reply on the crdt topic while joined falls through; the `!crdtJoined` aliasing window self-heals via the #191 onclose reset (documented inline).

v1.12.23. Full suite (2044 tests), build, biome ci, eslint, stylelint green.

https://claude.ai/code/session_01YPsYQart6AXhihbSjaqTqV


Rasbandit:
Paired backend PR engram-app/Engram#1008 now carries the e2e regression proof: deterministic deaf-note stage (room kill + zeroed handshake budget) FAILS on plugin main with the exact incident signature and PASSES on this branch in 41s. Merge this PR first; #1008's e2e hard-gates live-bound convergence from then on.

PR #248, opened 2026-07-15
fix(sync): bound request deadlines + post-pull push deferral

### LINKED ISSUE #244: Push deferral behind long post-swap pull chain stalls fresh notes 60s+ (was: mint queue flush)
## Evidence (backend CI run 29388970476, e2e-clerk, gw1 instance A)

Six tests failed in one run with push-direction timeouts (test_09/10/11/23/81/83). Not lost, STALLED:

- 04:39:21.5 socket joins: sync + user + crdt all JOINED, plan state received 04:39:22.6 (channel fully healthy)
- 04:39:23.4 client:vault create FromA-1/2/3 (test_09), routed `crdt=true confirmed=false live=false`
- ~68s of silence: no push, no channel traffic for these notes
- 04:40:31-32 notes materialize server-side, WS upsert fan-out echoes back (`Echo skip (pushing)`)
- 04:40:36 legacy `Push start` fires and echo-skips (hash already on server)

test_09 (15s wait), test_10 (30s), test_11 (10s), test_23 (10s) all expired inside the stall window. test_83 later failed with `connId:null` channel state on the same worker.

The ~68s delay is consistent with a periodic queue-flush timer (~60s) picking up notes that were minted immediately after a (re)join, i.e. no flush-on-join / flush-on-mint-confirmation trigger. The wedge began right after test_48 (OAuth swap + reconnect x2) finished on the same instance, suggesting the swap left the mint-confirmation or queue-drain trigger unarmed.

Same signature as earlier singleton failures: test_04 (`Version 2 not on server after 10s`, run 29207718756) and test_08 (run 29244132332). With the three big flake classes fixed (backend PR 1015 / plugin PR 243), this stall is likely the dominant remaining e2e-clerk flake.

## Suggested direction
Trace why a note created with `crdt=true confirmed=false live=false` right after a join waits for the periodic flush instead of an immediate drain; add a flush trigger on crdtJoined and on mint, or shorten the timer. May be subsumed by the vault-channel CRDT fan-out work (feat/vault-channel-crdt-fanout).

Artifacts: workspace scratchpad e2e-fails/29388970476; grep docker-compose.log for FromA-1.

Rasbandit:
## Root cause (from CI run 29388970476 artifacts + prod 2026-07-15)

A pull wedged on a half-open connection (deploy drain, OAuth swap) hung forever: `requestUrl()` has no timeout, so `pulling` stayed true indefinitely. From that moment:

- every future `pull()` short-circuited on the re-entry guard → **receives stopped**
- every local edit sat in `pendingPostPullPushes`, drained only when the pull ends (never) → **pushes stopped**

Sync dead in both directions until plugin reload. In CI (instance A, test_09): FromA-1/2/3 created 04:39:23 during a wedged post-OAuth-swap pull, zero push activity for 68s, rescued only by the NEXT FullSync's `pushModifiedFiles` sweep at 04:40:31 (whose own `pull()` returned 0 instantly via the re-entry guard). This is also the live-prod "swap a file and sync stops both ways, reload Obsidian fixes it" symptom.

## Fix — bound both sides of the wedge

1. **api.ts**: every request races a deadline (30s notes/meta, 120s attachments). `requestUrl` cannot abort, so the zombie is abandoned (late rejection swallowed; a late server-side apply is handled by the existing echo/conflict machinery). `Promise.race` (not re-wrap) so a real rejection propagates unchanged — callers classify on `.status` (402 limit parse, 401 token retry). A timeout carries no status, so existing callers treat it as connection loss: `pull()`'s `finally` drains deferred pushes and re-runs via `pullPending`; `pushFile` enqueues a durable retry.
2. **sync.ts**: `pendingPostPullPushes` now drains at most `postPullMaxDeferMs` (5s) after the first deferral even while a pull still runs. Safe: the deferral is an echo-traffic optimization, not a correctness gate — `pushFile`'s echo-hash gate filters sync-write echoes either way (the normal end-of-pull drain relies on exactly the same property).

## Tests

- `tests/api.test.ts`: never-settling request rejects with `RequestTimeoutError`; attachment paths use the longer deadline; fast responses unaffected; `health()` bounded (a wedged probe froze offline recovery).
- `tests/sync-postpull-defer.test.ts`: an edit deferred behind a wedged pull pushes within the defer bound while the pull is still pending; `destroy()` cancels the pending drain.
- Full suite: 2077 pass / 0 fail. biome + eslint(obsidian) + stylelint + tsc green.

Closes #244

https://claude.ai/code/session_019omnopWqAww7rfxG9Yp47i

## Also in this PR: fast wedge detection + evidence-based deadlines

Waiting out the full deadline is the backstop, not the recovery path. The channel disconnect is the earliest wedge signal the plugin gets (a file-swap pull rides the same dying connection as the dropped socket), so `onStatusChange(false)` now calls `EngramApi.failWedgedRequests()`: a fresh `/health` probe races any in-flight request older than 4s, and if the probe answers while they still hang, they are failed immediately with `RequestTimeoutError` — recovery in ~probe time instead of 15-30s. If the probe fails (real outage), requests keep their normal deadline.

Deadlines tightened: notes/metadata 30s → 15s (prod p99 is well under 2s); only actual byte transfers (`POST /attachments` upload, `GET /attachments/<path>` download) keep 120s — attachment DELETE and `/attachments/changes` are metadata and take the short deadline. Server-side complement (drain sends `connection: close`) ships in Engram#1018. New tests: wedge-probe reject / real-outage no-op / young-request no-op / metadata-deadline scope. Suite 2083/0.


Rasbandit:
Review wave applied (commit b50c745): anti-stale version guard in applyChange closes the mid-pull drain overwrite race (CONFIRMED data-loss finding); wedge probe now single-flight, shields attachment transfers, and deletes-before-abandon; /attachments/changes matched exactly (changes.pdf gets the transfer deadline); bulk pages get a 60s middle deadline; dead zombie-rejection swallows removed. Plugin suite 2091/0, all lints + build green.

PR #277, opened 2026-07-21
feat(sync): live-op behind-detector heals seq gaps via socket catch-up (single-path D2)

Rasbandit:
Phase D2 of the single-path CRDT migration (plugin half; pairs with the backend PR of the same branch name — merge backend first).

## What
- `SyncEngine.applyLiveOpWithSeq(noteId, seq, apply)`: a pure behind-detector wrapping every inbound `note_yjs_update` apply. The apply runs unconditionally in every branch (Yjs ops are commutative + idempotent; seq NEVER gates application). An integer `seq > catchupSeq` fires the existing single-flight `catchupViaSeqReplay()` and logs a `gap-heal fired` breadcrumb (noteId + seq + cursor) to the client-log pipeline.
- Live ops never advance the cursor: the seq replay is the sole cursor writer, now with a monotonic per-page guard (the cross-vault reset path that legitimately lowers the cursor is preserved).
- Runtime `Number.isInteger` guard at the channel frame boundary — string/NaN/float/null/absent seq all degrade to today's behavior byte-for-byte (old-backend back-compat).

## Why this shape
A review pass proved that advancing the cursor from live-op observation is unsound: fire-and-forget applies can silently skip (unmapped notes), a delta's stamped seq aliases missed renames/deletes (Note.seq comes from ANY write kind), and +1 arithmetic over the unordered cross-node multi-note stream false-gaps continuously. The behind-detector sidesteps all three: any observed seq > cursor is by construction a true "feed has unconsumed entries" signal, and the single-flight replay consumes them properly.

Gated by the backend's new deterministic e2e (`test_seq_gap_heal.py`): a dropped live broadcast now self-heals over the socket within the live window — the prerequisite for Phase E deleting the REST recovery path.

## Testing
- Full `bun test` 2168 pass / 0 fail; tsc + biome ci clean.
- New `tests/sync-gap-heal.test.ts`: every branch asserts the apply-exactly-once invariant + cursor state (fresh/gap/stale/null/undefined/string/NaN/float/fresh-install).
- `tests/channel-crdt.test.ts`: frame-boundary seq threading + non-integer normalization.

Version: 1.13.4.

https://claude.ai/code/session_01VK9qUcRtS3tiW9wNhw83y7


Rasbandit:
Superseded by #278 (merged), which ships the same seq gap-heal wire contract. This branch's adversarial-review hardening (behind-detector defect classes, snapshot-stomp Yjs-delta converge, null/NaN frame guards) is being audited against the merged implementation; any gaps land as a small follow-up PR. Branch retained for reference.

PR #280, opened 2026-07-22
fix(sync): seq fence strict — an equal-seq row can carry a server merge the device never saw

Rasbandit:
Correction to #278's per-path seq fence, caught by main run 29881173619 (`test_85_missed_delivery_no_deletion`).

**Failure:** device B pushed a local edit while a server edit sat undelivered. The server merged both into the row under the same seq B had just observed via its own push echo, so the `<=` fence classified the row as history (`stale row (seq 89/89) — history, skip`) and B never converged to the merged content.

**Fix:** strict `<`. Only rows provably *behind* the per-path high-water are history. The truly-stale equal-seq case (checkpoint-lagged projection of the op just applied) is already covered by the upstream `serverHash` equality gate, so the revert-ping-pong protection from #278 is preserved (pre-merge storm rows sit strictly below the high-water).

This ambiguity (seq equality cannot distinguish "lagged projection" from "merge containing unseen edits") is inherent to snapshot backfill — the proper fix is Phase E's routing of CRDT rows through the Yjs delta converge (`docs/context/d2-seq-replay-stale-snapshot-stomp.md`, direction 1).

Gate: tsc + biome + 2169 tests green. No version bump (release-please).

https://claude.ai/code/session_01KJksd3C2xLkXyucHRRnuVC

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -190,7 +190,9 @@ export class OAuthAuth implements AuthProvider {
 		// replaying it: the server may have already committed that rotation,
 		// and a replay of a consumed token 401s definitively (forced re-link).
 		const raw =
-			this.lateRefresh?.token === sentToken ? this.lateRefresh.raw : this.refreshFn(sentToken);
+			this.lateRefresh?.token === sentToken
+				? this.lateRefresh.raw
+				: this.refreshFn(sentToken);
 		try {
 			const result = await withTimeout(raw, OAuthAuth.REFRESH_DEADLINE_MS);
 			this.lateRefresh = null;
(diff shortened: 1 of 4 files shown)


Rasbandit:
Superseded by #296, which lands the fence fix in its correct **content-hash-aware** form. Root cause (confirmed from the backend's own `crdt_persistence.ex` GUARANTEE BOUNDARY): seq is shared across live updates, so a blunt strict-`<` would wrongly stop skipping genuine duplicate/echo rows. #296 instead makes an equal-seq row stale only when its content_hash matches — fixing the data-loss while preserving duplicate-skip. NOTE: this PR also bundled an unrelated OAuth refresh-rotation fix (auth.ts) — that's still worth landing; please split it into its own PR.

Issue #282, opened 2026-07-22
Seq/version fence skips feed heal when local push collides versions (test_85)


Rasbandit:
Found while triaging e2e for PR #281's paired run (run 29883968835, e2e-clerk).

**Evidence (client logs, docker-compose.log):** `[client:pull] applyChange skip (stale v3 <= synced v3): E2E/MissedDelivery-*/Note.md` — after a missed live delivery, the device's LOCAL push records v3 into syncState; the subsequent feed entry for the missed remote content also carries v3, and the staleness fence (merged in #278) skips it as `v3 <= v3`. The missed remote content is never healed via the feed and `test_85_missed_delivery_no_deletion` times out.

**Why it's wrong:** version equality does not imply content convergence. When a delivery was missed and the local push collided the version counter, the fence's `<=` comparison masks the exact heal the gap-heal feature exists to perform. The fence needs a content/head disambiguator for the equal-version case (or should pass equal-version entries through to the Yjs converge, which is idempotent and safe — see PR #281, which makes that pass-through harmless).

Rotating member of the endemic e2e-clerk pool; failed 2/3 rolls on run 29883968835 with identical signature.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01VK9qUcRtS3tiW9wNhw83y7

Rasbandit:
Confirming evidence from a second run — this time on **main WITH #281 merged** (run [29895069823](https://github.com/engram-app/Engram/actions/runs/29895069823), backend main, plugin main@fc44075): the class is NOT retired by the delta-converge port, and it is not ambient latency.

**Artifact evidence (`ci-debug-8fa9f1be…/docker-compose.log`):**

- 05:58:03, device `51112f6f` (B, the device that made the offline local edit): `[client:pull] CRDT catch-up: stale row (seq 137/137 v3/-) — history, skip E2E/MissedDelivery-349b9dddeb4e/Note.md` — an **equal-seq** row (137/137) skipped as history. Equal-seq rows can carry server merges of unseen edits; this one did (server content was `base + local-edit + server-edit`).
- **Zero `via Yjs converge` / backfill lines fire for this note anywhere in the run** — the #281 converge never ran on the reconnect catch-up path after the skip, so nothing healed it.
- Outcome: B stayed at `base + local-edit` through teardown (test assert: "B never converged to the server content after reconnect").

[...]

PR #286, opened 2026-07-22
feat(settings): adopt getSettingDefinitions (Obsidian 1.13 search)

### LINKED ISSUE #270: Adopt the Obsidian 1.13 declarative settings API (getSettingDefinitions)
## Why

Obsidian's community scanner and `eslint-plugin-obsidianmd` 0.4.x both report:

> This `PluginSettingTab` does not implement `getSettingDefinitions()`; its settings will not appear in Obsidian's settings search for users on 1.13.0 or later.

Our settings are invisible to Obsidian's global settings search on 1.13+.

## This is a full settings-UI rewrite, not an additive change

From the shipped typings (`obsidian.d.ts`, `SettingTab.display`):

> "Not called when `getSettingDefinitions` returns a non-empty array; the tab is rendered declaratively from those definitions instead."
> `@deprecated Since 1.13.0. Use getSettingDefinitions() instead.`

Confirmed by the [official migration guide](https://docs.obsidian.md/plugins/guides/migrate-declarative-settings). Returning a non-empty array means Obsidian 1.13+ renders the whole tab and **`display()` is never called**. Our tab bar, status bar, progress bar, device-flow auth, network-sourced vault dropdown and waitlist mini-app all live inside `display()`.

`return []` is the documented escape valve (leaves `display()` in charge) and silences the rule, since the lint rule is a pure syntactic presence check that never inspects the return value. That would be suppression wearing adoption's clothes, so it is deliberately not what we did.

## Hazards found during research

1. **`setControlValue`'s default bypasses our side effects.** It writes `data.json` and persists, but our `saveSettings()` also calls `api.updateConfig`, `syncEngine.updateSettings`, `rlog().setEnabled`, restarts the sync interval and re-registers the vault. Without an override, toggling Diagnostics would appear to save and do nothing until restart.
2. **`SettingPage` and `SettingGroup` are runtime classes.** Referencing one at module scope throws `extends undefined` on Obsidian < 1.13 and **bricks the plugin at load**. Invisible to `tsc`. Our `minAppVersion` is 1.7.2, so this would hit nearly every user.
3. **`openSyncCenterSettings()` would regress.** It deep-links to the Sync Center after sync errors via `setInitialTab` + `openTabById`. There is no public API to navigate to a declarative *page*, so that error-recovery flow would land on the tab root.
4. **Path B means two implementations.** 1.13 is still an insider build, so `display()` must stay as the fallback and the two must be kept in sync.

## Shape of the work

- 5 tabs map to `SettingDefinitionPage`, built lazily inside `page: () => ...` factories (hazard 2)
- 4 genuinely bindable settings become `SettingDefinitionControl`: `conflictResolution`, `debounceMs`, `ignorePatterns`, `diagnosticsEnabled`, plus a `setControlValue` override routing through `saveSettings()`
- Everything else (auth, vault dropdown, URL preflight, progress bar, waitlist) stays imperative behind `SettingDefinitionRender`
- Resolve the Sync Center deep-link question first; if pages cannot be navigated to programmatically, keep Sync Center at top level rather than shipping a worse error-recovery flow

## Smoke plan

1. Load on Obsidian 1.12.x or older, confirm no console error at load (catches hazard 2)
2. On 1.13.x, every control persists **and takes effect live** (toggle Diagnostics, confirm remote logging flips)
3. Settings search finds Diagnostics / Conflict resolution / Debounce / Ignore patterns
4. Device-flow sign-in, sign-out, API key, vault dropdown, URL preflight
5. Trigger a sync error, verify where `openSyncCenterSettings()` lands

## Meanwhile

`obsidianmd/settings-tab/prefer-setting-definitions` is disabled in `eslint.config.mts` with a pointer to this issue. The scanner will keep reporting it as one warning until this lands.

Rasbandit:
## Summary

- Adopt Obsidian 1.13's declarative settings API: `getSettingDefinitions()` returns a single `render`-hatch item that hosts the existing custom settings UI, so the tab appears in Obsidian 1.13+ settings search. `display()` is kept as the pre-1.13 fallback (minAppVersion is 1.7.2); both call the extracted `renderContent()`.
- We do **not** decompose settings into declarative `control` objects — the tab is a rich custom UI (tab bar, status, progress, device-flow) with no 1:1 declarative form, and a non-empty return replaces `display()` entirely on 1.13+. The render-hatch is additive and loses nothing. Ceiling: indexed as one search entry, not per-setting.
- Enforce `obsidianmd/settings-tab/prefer-setting-definitions` (was `off`, #270).
- `main.ts` refreshes the tab via a new public `rerender()` instead of the now-deprecated `display()`, which on 1.13+ would draw into the tab root rather than the render-hatch host.

Obsidian typings are intentionally held at 1.8.7: bumping to 1.13 deprecates `setWarning()` in favor of `setDestructive()` (1.13-only), which the lint policy can neither use nor suppress at minAppVersion 1.7.2. The 1.13 API is typed against a local shim instead. No version bump (release-please owns it).

## Test plan

- [x] `tsc`, `lint:obsidian` (`--max-warnings 0`, rule enforced), `biome`, `stylelint` all green
- [x] `bun test` — 2172 pass / 0 fail
- [x] `bun run build` clean; bundle contains `getSettingDefinitions` + `engram-settings-host`
- [x] Negative proof: removing the method makes the rule fire with the documented message; restoring passes
- [ ] **Manual (needs Obsidian 1.13, currently Catalyst-only):** open Settings → Engram, confirm the tab renders edge-to-edge and surfaces in settings-search
- [ ] **Manual (1.12.7):** confirm the `display()` fallback renders identically (the 1.13 render-hatch path is dormant on current stable)

Closes #270

https://claude.ai/code/session_011kosypXtQRFzj9gvowKKgS


Rasbandit:
backend/e2e red triaged before rerun: e2e-crdt failed on test_deaf_live_bound_note_converges_via_rest_catchup (deaf live-bound class, being deleted/fixed by single-path D3, plugin PR #284 + backend #1066 latency work) and e2e-clerk failed test_48 x2 / test_27 / test_30 — all delivery-timeout classes tracked in plugin #282/#283 and backend #1066. Same classes fail on backend main runs with no plugin change (main CI 29897680251 vs green 29897325621 six minutes apart). This PR touches only the settings tab, eslint config, and the built main.js — no sync path. Rerolling the failed jobs with the same pinned PLUGIN_SHA.

Rasbandit:
### Code review

Found 1 issue:

1. `rerender()` checks `activeContainerEl` for null but not for staleness (`isConnected`). On the 1.13+ `getSettingDefinitions()` path, `activeContainerEl` is a div inside the declarative row (`setting.settingEl.createDiv()`), which Obsidian can tear down and recreate on settings re-renders/tab switches without ever calling the tab's `hide()` — so a late async `rerender()` (device-flow OAuth completion in `startDeviceFlow()`, or the vault-switch refresh at `src/main.ts:2239`) renders into a detached node and the UI silently never updates. This regresses the class of bug commit 126f78c fixed for `renderStatus()` in this same file, which guards with `if (!statusEl || !statusEl.isConnected) return;` (line 238). Fix: `if (this.activeContainerEl?.isConnected) this.renderContent(this.activeContainerEl);`

https://github.com/engram-app/Engram-obsidian/blob/57edcd88724210c9512cb26c93259975c5f3223c/src/settings.ts#L89-L93

Rasbandit:
Review finding addressed: `rerender()` now guards with `activeContainerEl?.isConnected` (mirrors `renderStatus()`'s 126f78c pattern). Full local gauntlet green: build, 2173 tests, eslint --max-warnings 0, biome ci, lint:obsidian, lint:css.

PR #291, opened 2026-07-22
feat(sync): seq-diff vault validator — integer diff, bounded re-serve, head-equality heal skip (Phase E1)

Rasbandit:
Phase E1 of the single-path CRDT migration — the plugin half of the seq-diff bulk vault validator (Engram#1065, LOCKED addition). Pairs backend branch `feat/crdt-manifest-seq-validator` (same name — cross-repo e2e pairs on it).

## What
- **`validateFromManifest`** — pure integer diff: flags a manifest note row whose `seq` the replay already CONSUMED (`row.seq <= catchupSeq`) but the path never recorded — the silent apply-loss class behind e2e test_10's `received=yes materialized=no` — and rewinds the cursor to `min(behind)-1` so ONE replay re-serves everything. Rows beyond the cursor are untouched (the imminent replay fetches them anyway). Bounded: one rewind per distinct discrepancy per session — no rewind loops.
- **`healDivergedLiveBoundNotes`** — skips on `crdt_head` equality: op-level proof the doc holds the server state, terminating the D3 STEP1-refire ceiling for open notes whose serverHash never recorded (the ponytail ceiling noted in #284).
- **`catchUp`** — sends `?since_seq=<manifestSeq>`; an `unchanged` response skips every manifest-driven step (reconcile, validator, heal, folder sync — every server write bumps `change_seq`) while the seq replay still runs on its own cursor. The watermark records only after a fully-completed pass and is wiped with per-vault state (cross-vault integer-collision guard).
- Types: `ManifestEntry.seq/crdt_head`, `ManifestResponse.unchanged`; `getManifest(sinceSeq?)`.

## Why not hashes
`content_hash` is a per-user keyed HMAC (uncomputable client-side) and equality is direction-blind (the D2 stomp class) — decision locked 2026-07-22, see the Phase E plan + #1065.

## Tests
10 new unit tests (validator diff cases incl. legacy-entry/old-backend/bounded-retry, head-equality skip both ways, catchUp short-circuit + watermark recording). Full suite 2204 pass / 0 fail; build + biome + eslint + stylelint green. No version bumps (release-please).

### DIFF
--- a/main.js
+++ b/main.js
@@ -18167,8 +18167,15 @@ var BINARY_EXTENSIONS = /* @__PURE__ */ new Set([
      *  Persisted under `manifestSeq`; wiped with the per-vault state. */
     this.manifestSeq = 0;
     /** Cursor value of the last validator rewind — bounds the validator to ONE
-     *  re-serve per distinct discrepancy per session (see validateFromManifest). */
-    this.lastValidatorRewind = -1;
+     *  re-serve per distinct discrepancy per session (see validateFromManifest).
+     *  null = no rewind yet (a numeric sentinel would collide with the
+     *  legitimate `minBehind - 1` target domain, which includes -1 and 0). */
+    this.lastValidatorRewind = null;
+    /** A pending validator rewind, consumed atomically by the next
+     *  `runSeqReplayOnce` (`catchupViaSeqReplay` is the SOLE cursor writer —
+     *  a direct `setCatchupSeq` here would race an in-flight replay's per-page
+     *  cursor persist and be silently clobbered). */
+    this.seqRewindFloor = null;
     /** When true, vault delete events are suppressed (used during local wipe). */
     this.suppressDeletes = !1;
     /** Paths modified during a pull that need pushing once pull completes. */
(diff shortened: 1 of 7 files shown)


Rasbandit:
### Code review

Found 3 issues (all fixed in af3577c before merge):

1. The `unchanged` short-circuit skipped `seedEmptyFolders` and could strand a fire-and-forget heal forever: the watermark recorded even when a live-bound heal had only just been POKED, so an idle vault short-circuited every later poll and the retry net died. Fixed: the local-to-server folder seed always runs, and the watermark records only on a fully-clean pass (zero behind rows, zero heal pokes).

https://github.com/engram-app/Engram-obsidian/blob/af3577caf99cfeaf76954c2bf427b446d7466150/src/sync.ts#L3486-L3497

https://github.com/engram-app/Engram-obsidian/blob/af3577caf99cfeaf76954c2bf427b446d7466150/src/sync.ts#L3511-L3519

2. `validateFromManifest` wrote the seq cursor directly, racing an in-flight `runSeqReplayOnce` whose per-page persist would silently clobber the rewind while `lastValidatorRewind` latched (defeating the one-shot re-serve). Fixed: the validator hands a `seqRewindFloor` to the replay, which consumes it atomically at run start — `catchupViaSeqReplay` stays the sole cursor writer per its documented invariant.

https://github.com/engram-app/Engram-obsidian/blob/af3577caf99cfeaf76954c2bf427b446d7466150/src/sync.ts#L5093-L5099

3. The plain apply branches (identical/overwrite/create) never stamped `seq` into `syncState`, so a previously-synced non-CRDT note kept `seq: undefined` forever and the validator's legacy-entry rule made it permanently unflaggable for later-update losses. Fixed: all three sites record `change.seq` (preserving prior values for legacy changes without one).

https://github.com/engram-app/Engram-obsidian/blob/af3577caf99cfeaf76954c2bf427b446d7466150/src/sync.ts#L5763-L5771

Issue #234, opened 2026-07-11
CRDT: history-less doc baseline-doubling when capturing disk drift before a remote apply


Rasbandit:
## Summary

Follow-up from the vault-channel fan-out PR #233 (adversarial re-review of the P4 data-loss fix).

`captureDiskDriftBeforeRemote` (src/sync.ts) seeds an un-pushed on-disk edit into the Y.Doc via `applyLocalEdit` **before** applying a fanned-out/cold-received remote update, so CRDT merges the local drift instead of the remote projection clobbering it. For a **history-full** doc this is a clean 3-way merge.

For a **history-less** doc, it doubles the shared baseline.

## Scenario

1. Idle note N is confirmed and written to disk via the cursor feed (`applyChange`) — this does NOT open N's Y.Doc, so its IndexedDB store is history-less. (Or IDB was evicted, e.g. iOS.)
2. N is not live-bound. User edits N on disk; the debounced `pushFile` hasn't fired.
3. Another device edits N → a `note_yjs_update` fan-out (or a `coldReceive` poll) lands.
4. `captureDiskDriftBeforeRemote` → `applyLocalEdit(N, disk)`: doc empty, `lca=false`, `isUnchangedSynced` false → `seedOnce` inserts the **entire disk (baseline + local edit)** as a fresh lineage.
5. `applyRemoteUpdate` merges the server lineage (also containing the baseline). Yjs unions two independent insertions of the baseline → **baseline text DOUBLES**. Both edits survive; the baseline is duplicated, permanently.

## Severity / scope

- **Not a blocker for #233.** It replaces a *silent loss* (BUG 2) with recoverable *doubling* — strictly better.
- **Pre-existing seedOnce behavior:** the legacy `pushFile` path doubles a diverged history-less confirmed note the same way (server doubles on receive). `captureDiskDriftBeforeRemote` adds a second trigger on the receive path; it does not create a new doubling class.
- **Bounded window:** history-less + un-pushed local drift + first push, all at once. Once a note is history-full (any prior push/STEP2), the merge is clean.
- **Permanent once it happens** (a single doubling won't hit the 1000-client flatten threshold) — so "self-healing" is inaccurate.

## Proper fix

Needs a real 3-way / LCA merge for **history-less** docs. Naive order-flips do NOT work:
- seed-disk-first (current) → doubles the baseline.
[...]

PR #293, opened 2026-07-22
feat: delete the REST Yjs transport — socket is the only delta path (Phase E3)

Rasbandit:
Phase E3 of the single-path CRDT migration (plan: workspace `docs/superpowers/plans/2026-07-22-crdt-single-path-phase-e.md`). Paired backend PR: engram-app/Engram#1072 (same branch name — e2e auto-pairing; merge backend first, then re-trigger here).

## What

Every remaining REST Yjs-delta path is deleted; the `crdt:` socket sv-exchange is the only delivery mechanism for CRDT ops.

- **Stream events never fetch**: a hash-only / meta-projected / distrusted-`""` upsert routes to the op-log seq-replay (rows carry the real content) instead of a `getNote` fetch. The empty-content hash is taught by authoritative op-log rows, not a learn-by-fetch.
- **Cold catch-up leg**: `restConvergeAndFlush` replaced with the SAME stage-then-fire pattern the live-bound leg uses (stage `pendingConvergence` → `socketConverge` → content-verified `commitCrdtConvergence`). This includes the disk-text-equals-row-text case — text equality is never treated as op-level proof.
- **Gap heal** (`applyPushedNoteUpdate`): fires the cooldown-gated socket re-handshake instead of a REST full-delta pull; `crdtHead` stays unadvanced until the doc actually reaches a head.
- **History-less adopt**: refires the enroll (empty-sv STEP1 → server sends full state). Any disk drift is preserved to a keep-both conflict copy BEFORE the refire, and the original's baseline advances so drift can't seed a fresh lineage (the #234 doubling door). **Named tradeoff:** the LCA 3-way auto-merge for history-less drift (`reconcileDriftOntoServer`) is deleted — that rare compound case now always surfaces a keep-both conflict copy (data-safe: both versions survive) because STEP2 full state arrives asynchronously with no post-adopt merge hook.
- **Offline queue drain**: a durable crdt entry fires reset+enroll when the channel is live — the server answers the client STEP1 with `[STEP2, server-STEP1]` (verified in the y_ex NIF), and the client's reply carries the pending local ops. The entry is NOT dequeued on fire: it settles in `commitCrdtConvergence` once an inbound frame proves the round-trip; a lost nudge re-fires next flush. Channel down → entry stays queued; there is no REST side-channel anymore.
- **Deleted**: `api.getUpdates`, `api.postUpdate`, `restConvergeCore`, `restConvergeAndFlush`, `reconcileDriftOntoServer`; `socketConvergeLiveBound` renamed `socketConverge` (e2e-grepped log strings unchanged).

## Safety invariants kept

- Heal converges via Yjs deltas, never content snapshots (D2 stomp class).
- Nothing records convergence or dequeues a delivery before op-level proof (content-verified commit / inbound-frame settle).
- Drift is never silently lost: keep-both copy before any flow that lets STEP2 overwrite disk; copy-write failure aborts without refiring.

## Review

5-agent review ran on 42b378a; 3 confirmed findings (queue dequeue-before-proof, text-equality fast path, biome CI red) all fixed in b886209 — see the review comment. Stale docstrings + `docs/context/sync-catchup-convergence.md` refreshed in the same commit.

## Gates (at b886209)

- `bun test`: 2197 pass / 0 fail
- `bun run build`, `bunx biome ci`, `lint:obsidian`, `lint:css` clean
- grep-prove: zero `getUpdates`/`postUpdate`/`restConverge` references in `src/`

### DIFF
--- a/docs/context/sync-catchup-convergence.md
+++ b/docs/context/sync-catchup-convergence.md
@@ -1,11 +1,13 @@
 # Context Doc: Catch-up convergence (missed-delivery healing)
 
-_Last verified: 2026-07-08_
+_Last verified: 2026-07-22 (Phase E3 update)_
 
 ## Status
 Working — shipped in plugin PRs #197 (v1.11.22) + #198 (v1.11.23), backend v0.5.642; hardened by #207/#209/#211 (v1.11.31-33) after reruns=0 unmasked three holes in the original design (see "The four guards").
 
-**Superseded (2026-07, Plan B1 CRDT-authoritative rewire, Task 6):** the per-open handshake `verifyConvergenceOnOpen` described in (c) below was REMOVED. File-open is now a pure local bind (`this.crdtLiveViews?.refresh()`, no hash comparison, no REST round-trip). Convergence on missed deliveries is now owned entirely by socket vault-catchup (`SyncEngine.catchupViaSocket()`) run at connect/reconnect (`main.ts` `onStatusChange`) and at `onLayoutReady`, over `crdt_catchup_heads`/`crdt_catchup_delta`. The rest of this doc (id adoption, base_hash CAS, backfill, the four guards) is still accurate — only mechanism (c) changed.
+**Superseded (2026-07, Plan B1 CRDT-authoritative rewire, Task 6):** the per-open handshake `verifyConvergenceOnOpen` described in (c) below was REMOVED. File-open is now a pure local bind (`this.crdtLiveViews?.refresh()`, no hash comparison, no REST round-trip). Convergence on missed deliveries is now owned entirely by socket vault-catchup (`SyncEngine.catchupViaSocket()`) run at connect/reconnect (`main.ts` `onStatusChange`) and at `onLayoutReady`, over `crdt_catchup_heads`/`crdt_catchup_delta`.
+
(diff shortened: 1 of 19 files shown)


Rasbandit:
### Code review

Found 3 issues (all addressed in b886209 on this branch):

1. The offline-queue drain dequeued a durable crdt entry immediately after the fire-and-forget `socketConverge` nudge — no delivery proof, and `CrdtEnrollment.enroll()` surfaces no failure, so a STEP1 lost to a socket drop stranded the edit (durable in the Y.Doc but with nothing left to re-trigger delivery until the note was next opened). This contradicted the branch's own "never record before op-level proof" principle. Fixed: the entry now settles in `commitCrdtConvergence` when an inbound frame for the note proves the room round-tripped; a lost nudge re-fires on the next flush (cooldown-bounded).

https://github.com/engram-app/Engram-obsidian/blob/42b378affaf8e6ad94c209c259eda2a6138812bc/src/sync.ts#L7082-L7091

2. The new cold-leg "quietly converged" fast path recorded `serverHash`/`version`/`seq` from disk-vs-row TEXT equality with no Yjs round-trip — the exact "text equality is not op-level proof" class the D3 redesign closed for the live-bound leg (two independently-typed identical bodies are a disjoint lineage → #234 doubling), and it advanced the seq fence without ops, worsening the open #282 class. Fixed: the fast path is deleted; the case routes through the same stage-then-fire + content-verified commit as every other leg.

https://github.com/engram-app/Engram-obsidian/blob/42b378affaf8e6ad94c209c259eda2a6138812bc/src/sync.ts#L5236-L5249

3. CI `build-and-test` was red on `bunx biome check` (unformatted `spyOn(...)` lines in `tests/sync-socket-catchup.test.ts`, fallout of the `socketConvergeLiveBound`→`socketConverge` rename sed), contradicting the PR body's "gates clean" claim. Fixed: formatted; `bunx biome ci src/ tests/` clean.

Also noted, accepted as a documented tradeoff (not changed): `adoptHistoryLessNote` downgrades the history-less-drift LCA 3-way merge to an unconditional keep-both conflict copy — data-safe (nothing lost, both versions survive), required because STEP2 full state now arrives asynchronously with no post-adopt merge hook. Stale docstrings referencing the deleted REST machinery were refreshed in the same commit, and `docs/context/sync-catchup-convergence.md` gained a Phase E3 supersession stanza.

PR #311, opened 2026-07-23
feat: sync preview on the op-log socket feed; delete /notes/changes + /attachments/changes client path

### LINKED ISSUE #304: Port sync-preview (computeSyncPlan) off REST getChanges/getAttachmentChanges to enumerate-only socket replay
From the CRDT audit: computeSyncPlan (sync.ts) is the LAST caller of GET /notes/changes + GET /attachments/changes. pushAll already enumerates the identical server set via catchupViaSeqReplayExclusive({fromZero, enumerateOnly}) over crdt_catchup_since. Extend enumerate-only to return paths/deleted flags, feed the preview modal from it, then the backend can delete both endpoints. This is the unfinished half of Phase E Half 2 (engram#1031).

Rasbandit:
Closes #304

## What

- **`enumerateServerState()`** — walks `crdtCatchupSince` from cursor 0 (500/page), folds last-seq-wins per note id / attachment path into authoritative server state. Requires the live socket (throws otherwise); never touches the real catch-up cursor.
- **`computeSyncPlan`** rewritten on that single fold — the REST-era manifest-fetch + `GET /notes/changes` delta + epoch-widening split is gone. Per-path classification: tombstone → delete-local, identical bytes → clean, server-unchanged (row hash == recorded serverHash) + local-changed → push, both-moved → conflict, else → pull. Tombstoned paths are never pushed (delete-wins).
- **Endpoint deletion** — `api.getChanges` / `api.getAttachmentChanges`, `ChangesResponse` / `AttachmentChangesResponse`, `fetchAllNoteChanges`, and the whole bulk deadline class (`bulkTimeoutMs`) — no caller of `/notes/batch`, `/notes/changes`, or `/sync/changes` remains, so deadline classification collapses to attachment-transfer vs default. Sim model-server `/notes/changes` + `/vault/heads` endpoints pruned too.
- **Tests** — 6 new enumerateServerState tests (fold, rename no-ghost, tombstones, pagination, not-live throw, cursor untouched); all 17 legacy computeSyncPlan tests ported to socket feed fixtures + new coverage (channel-down rejects, tombstone-never-pushed, serverUnchanged→push). Header vehicle tests (X-Vault-ID, X-Device-Id, GET tracing) rewired onto `getManifest`; 41 stale mock stubs stripped across harnesses.

## Backend follow-up

`GET /notes/changes` and `GET /attachments/changes` lose their last plugin caller. Backend-side deletion is **deploy-gated** like `/vault/heads` (Engram#1088) — old plugin versions in the field still call them until this release is out.

## Gauntlet

- `bun test`: 2252 pass / 0 fail
- `biome ci src tests`: clean
- `bun run build` (tsc + esbuild): clean

### DIFF
--- a/main.js
+++ b/main.js
@@ -1633,6 +1633,13 @@ var LARGE_FRAME_WARN_BYTES = 1e6, NoteChannel = class {
   get userTopic() {
     return `user:${this.userId}`;
   }
+  /** The server vault this channel is joined to (`crdt:{userId}:{vaultId}`),
+   *  or null before a vault is bound. Callers reading vault-scoped state off
+   *  the socket must confirm this matches their intended vault — a vault
+   *  switch that skips setupNoteStream leaves this channel on the old vault. */
+  getVaultId() {
+    return this.vaultId;
+  }
   get crdtTopic() {
     return this.vaultId ? `crdt:${this.userId}:${this.vaultId}` : null;
   }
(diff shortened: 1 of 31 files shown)


Rasbandit:
Addressed review in f6af43a:
- **Blocker (vault switch):** the catchup-since wiring closure now throws when the channel's bound vault ≠ `settings.vaultId` (new `CrdtChannel.getVaultId()`). `applyVaultChange` no longer renders the old vault's state as the new vault's plan.
- **Null-path fold guard:** `enumerateServerState` skips id-less or path-less rows (leaked folder-marker ops), mirroring `applyOp`. RED test added.
- **Stale comments:** reworded the deleted-endpoint doc comments to the op-log feed; dropped the dead `fetchAllNoteChanges`/`resolveChangeBody` block + unused `metaChange`.
- **Equal-seq attachment-move page split (minor, pre-existing + shared with the live replay):** filed as #312 rather than expanding this PR into the `crdt_catchup_since` frame contract.

PR #315, opened 2026-07-23
feat: thread + persist the composite {seq,id} catch-up cursor (#312)

### LINKED ISSUE #312: Op-log seq-replay drops one row of an equal-seq attachment-move pair split across a page boundary
## Problem

The seq-replay cursor (`runSeqReplayOnce` and the new `enumerateServerState` fold, #304) pages the op-log by a **seq-only** cursor: the channel reply carries `next_seq` and the next fetch queries strictly `seq > cursor` (backend `notes.ex` — `crdt_catchup_since`).

An attachment **move** writes two rows sharing **one** seq by design (#614, `attachments.ex`): the repoint at the new path + a tombstone at the old path. If a 500-row page boundary falls between them, the second row (old-path tombstone) has `seq == cursor` and is skipped by the strict `>`. Result: a ghost non-deleted old path survives in the replay / preview.

## Scope

Pre-existing in the live `runSeqReplayOnce` path; #304 added `enumerateServerState` as a second consumer whose doc comment asserts fold completeness, which surfaced it. Rare (needs a move to straddle an exact 500-row boundary) and shared with the live path, so not a regression — filing rather than expanding #304 into the frame contract.

## Fix options

1. Carry a composite `{seq, id}` cursor through the `crdt_catchup_since` frame (backend adds `after_id`; query becomes `(seq, id) > (cursor_seq, cursor_id)`). Correct + minimal rows.
2. Client re-fetches each page with a `>=` seq overlap and dedupes by `{seq, id}`. No backend change, one duplicate row per page.

Option 1 is the clean fix (touches backend + plugin). Found in PR #311 review.

Rasbandit:
Closes #312 (plugin half). Pairs with backend engram-app/engram#1097.

## Problem
The seq-replay pagination (`enumerateServerState` for the sync preview + `runSeqReplayOnce` for catch-up) sent `cursor_seq` only. An attachment move writes two rows at **one seq**; when a page boundary splits them, the backend's `seq > cursor` skips the second on the next fetch — a ghost old path survives.

## Fix
- `crdtCatchupSince` (channel + engine type) gains `cursorId` in / `next_id` out.
- New `cursorAdvances()` does the strict `(seq, id)` keyset comparison. Both loops advance on it, so `next_seq` **equalling** `cursor` with a greater id keeps paging — the old seq-only `next_seq > cursor` guard would have wrongly stopped at the pair boundary.
- `catchupId` is persisted alongside `catchupSeq` (new field + accessors, in the data.json blob, loaded on startup, reset on vault change / validator rewind) so an **interrupted** replay resumes at `(seq, id)` and can't skip the sibling. The hash-aware gap-heal fence still reads `catchupSeq` (seq) only — deliberately untouched.

**Backward-compatible:** a pre-#312 backend returns no `next_id` → `cursorId` stays null → seq-only, exactly as today. So this can merge/ship before or after the backend without breakage.

## Tests (TDD)
- `enumerateServerState`: asserts page 2 is fetched with the composite cursor `{5, "aaa"}` from page 1's `next_id` — verified RED (fails to `{5, null}`) without the `cursorId` send.
- `runSeqReplayOnce`: asserts the composite fetch on page 2 AND that `catchupId` persists to the last row's id.

## Gauntlet
`bun test` 2256 pass / 0 fail · `biome ci` clean · `lint:obsidian` clean · `bun run build` clean.

## Deploy note
The composite fix is only *active* once backend #1097 is in prod (it supplies `next_id` + honors `cursor_id`); until then the plugin runs seq-only as before. Not deploy-gated (degrades gracefully), but the pair should land together.

### DIFF
--- a/main.js
+++ b/main.js
@@ -1697,9 +1697,11 @@ var LARGE_FRAME_WARN_BYTES = 1e6, NoteChannel = class {
    *  merged notes+attachments feed), so it is causally complete and can
    *  never pend the way a state-vector delta can — this is what a
    *  reconnecting device replays to converge (deaf-note fix, e2e test_85). */
-  async crdtCatchupSince(cursorSeq, limit) {
-    let payload = { cursor_seq: cursorSeq };
-    return limit !== void 0 && (payload.limit = limit), await this.sendRequest("crdt_catchup_since", payload);
+  async crdtCatchupSince(cursorSeq, limit, cursorId) {
+    let payload = {
+      cursor_seq: cursorSeq
+    };
+    return limit !== void 0 && (payload.limit = limit), cursorId && (payload.cursor_id = cursorId), await this.sendRequest("crdt_catchup_since", payload);
   }
   async connect() {
     this.ws || (this.reconnectMs = 1e3, this.joinFailureBackoffMs = 1e3, await this.openSocket());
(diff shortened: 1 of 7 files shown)


Rasbandit:
Addressed the review BLOCKER in bf714d3:
- **Wiring dropped cursorId (blocker):** the `setupNoteStream` closure was arity-2 (`(cursorSeq, limit) => channel.crdtCatchupSince(cursorSeq, limit)`), so `cursor_id` was never forwarded in the real plugin — the whole #312 fix was inert in prod, and the unit tests missed it (they replace the closure with a 3-arg mock; TS bivariance accepts the shorter closure). Extracted the wiring into a tested helper `makeCrdtCatchupSender(channel, currentVaultId)` (keeps the #314 vault-mismatch guard, forwards all three args); `channel.test.ts` now asserts it forwards `cursorId` (verified RED without the forward) and throws on a vault mismatch.
- Clean on all other axes per the review (loop safety, UUID ordering, persistence consistency, fence untouched, backward-compat).
- Full gauntlet re-run green: 2259 tests, biome, obsidian-lint, build.

Issue #299, opened 2026-07-23
CRDT: pull-only rejoin handshake can permanently lose offline/held bound edits


Rasbandit:
Candidate data-loss defect surfaced (and faithfully reproduced) by the convergence sim in #298, held for real-backend confirmation before treating as a prod bug.

## Claim
A live-**bound** note edited while offline (or held by `canSendLive` before the note's `crdtHead` is set) can permanently lose those Y.Doc ops.

## Evidence (verified against source in #298 review)
- **Pull-only re-enroll:** on reconnect the client re-enrolls via STEP1 and the backend replies STEP2 only — it never STEP1s *back* to solicit the client's unsent ops (`src/sync.ts:786-790`, `flushHeldEditsOnCreateAck` at ~798 is a PULL, not a push). So ops that never reached the server live are not re-pushed on rejoin; they reach the server only on the note's NEXT online edit via a full-state flush.
- **No REST fallback for bound notes:** `handleModify` short-circuits the disk-driven CRDT route for a live-bound path (`src/sync.ts:1982` — `if (crdtManaged && this.isLiveBound(file.path)) return`), so the edit can't fall back to the legacy REST offline queue either.

## Sim repro
Any seed of the two-phase random workload with the offline/offline-online op enabled diverges (~100%): the offline replica's held edits stay local while all others converge. `model-server` replies STEP2-only on rejoin, mirroring the real backend. These ops are deliberately **scoped out** of the gated sim suite (not oracle-loosened) precisely because they expose this.

## Before filing as confirmed / fixing
Needs the **P2 real-backend headless tier** to isolate whether this is prod-real vs. a sim create/enroll-timing artifact. If confirmed: the fix is to have the client re-push held/uncommitted Y.Doc ops on rejoin (or let a bound note fall back to the REST queue when live delivery was never confirmed).

Related: #298 (sim), the testing-architecture migration.

PR #325, opened 2026-07-25
fix(crdt): recover offline edits after switch-away

Rasbandit:
## Problem

Prod report: *"moving between files and making edits, only some make it to the server."* Confirmed in Loki — `[client:channel] sendCrdt refused (crdt topic not joined): joined=false` on the user's device, across many notes, always preceded by a socket drop (`heartbeat unanswered` / `net::ERR_NETWORK_CHANGED`).

**Root cause:** an edit typed while the CRDT socket is down is refused by `channel.sendCrdt` (topic not joined) and dropped. The rejoin recovery (`reEnrollOpenCrdtNotes`) only re-enrolls notes **still open in an editor**, so an edit made to a note then **switched away from** is never re-solicited by the mutual handshake — stranded on local disk, never reaching the server. #299's recovery covers only still-open notes; this is the switch-away gap.

Reproduced deterministically on current `main` in the sim tier (faithful model server): #299 (note stays open) recovers, but close-before-reconnect strands the edit.

## Changes

**A — the fix.** Track docIds whose `sendCrdt` was refused while unjoined (shared `wiring.ts` send wrapper); on the `crdt:` rejoin, re-enroll them in addition to open leaves. `startSync` reopens the doc from IndexedDB, fires STEP1, and the mutual handshake re-solicits the held struct → converges. Cleared when a later send for the doc succeeds.

**B — lag.** One file switch fires `crdtLiveViews.refresh()` 2–3× (active-leaf-change + file-open + layout-change). Coalesce same-microtask duplicates — the first call runs synchronously as before; duplicates observing identical workspace state skip the O(open-leaves) rebind. Never defers or drops a refresh that sees new state.

**C — log noise.** Throttle the `sendCrdt refused` warn per doc (an offline burst logs once, not per frame) and reword it — with A the frame is recovered on rejoin, so it reads as a transient, not silent loss.

## Test

New differential sim gate **#299b switch-away** (`tests/sim/`): RED without `wiring.reEnrollUnsent()` on the rejoin path, GREEN with it. Asserts the edit reaches the **server** (a plain convergence check would false-green if reconnect *reverted* the edit).

- `bun run build` clean · **2227 pass / 1 skip / 0 fail** (incl. random fuzz sim) · biome + eslint clean · no version bumps.

## Not in this PR

- **Deeper lag** (scope `refresh()` to the focused leaf; reconsider per-open `healNoteOnOpen`) needs real-app profiling before touching this race-sensitive code — deferred.
- **The real Loki bloat** is the mTLS client cert on every shipped client log (`Engram.Logs`, backend repo) — separate PR.

### DIFF
--- a/docs/context/crdt-offline-edit-switch-away-dataloss.md
+++ b/docs/context/crdt-offline-edit-switch-away-dataloss.md
@@ -1,0 +1,47 @@
+# Context Doc: CRDT offline edit lost after switching away from the note
+
+_Last verified: 2026-07-24_
+
+## Status
+Fixed (branch `fix/crdt-offline-edit-lost-on-rejoin`, commit `ed9dfb5`)
+
+## Symptom
+In Obsidian, moving between files and making edits - only *some* edits reach the server. Prod Loki tripwire (`metadata_category=client`):
+
+```
+[client:channel] sendCrdt refused (crdt topic not joined): joined=false
+```
+
+Always preceded by `heartbeat unanswered - closing dead socket` or `net::ERR_NETWORK_CHANGED` - i.e. the socket died, an edit was typed during the dead window, then dropped.
+
+## Root Cause
+An edit typed while the CRDT socket is down is refused by `channel.sendCrdt` (`src/channel.ts:351`, gates on the vault-wide `crdtJoined` flag, which is `false` whenever the socket is closed) and dropped. `crdtJoined` is a SINGLE vault-wide flag (topic `crdt:{user}:{vault}`), not per-doc.
+
+The rejoin recovery `reEnrollOpenCrdtNotes` (`src/main.ts:1625`) only re-enrolls notes still open in an editor leaf. So an edit made to a note, then SWITCHED AWAY FROM (editor closed) before the socket reconnects, is never re-solicited by the mutual STEP1 handshake - it is stranded on local disk, never reaching the server.
+
+#299's recovery only covered still-OPEN notes; this is the **switch-away gap**.
+
+## Recovery mechanism (why re-enroll is what matters)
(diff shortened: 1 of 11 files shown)


Rasbandit:
### Code review

Ran a multi-agent review (5 independent reviewers: CLAUDE.md compliance, bug scan, git-history, prior-PR comments, code-comment invariants) followed by per-finding confidence scoring. No finding cleared the confidence bar to block. Two sub-threshold robustness items were worth hardening, and are addressed in `c3206f4`:

**1. `unsentDocIds` lacked the bounds its sibling `CrdtOpQueue` enforces** (no cap, no prune-on-delete). A note edited then deleted while offline kept its id in the set, so reconnect would re-fire a spurious STEP1 for a dead doc (racing delete-wins). Now capped at 500 with oldest-eviction, and pruned via `forgetUnsent` wired from the vault delete handler.

https://github.com/engram-app/Engram-obsidian/blob/c3206f45f14eb1f4dd1c99c0c0b1a133b69c328b/src/crdt/wiring.ts#L316-L333

https://github.com/engram-app/Engram-obsidian/blob/c3206f45f14eb1f4dd1c99c0c0b1a133b69c328b/src/main.ts#L633-L639

**2. `refresh()` coalescing was untested** despite this file's file-switch race history (#194). Added a regression test (same-tick duplicate skips the rebind pass; a later-tick call still runs).

https://github.com/engram-app/Engram-obsidian/blob/c3206f45f14eb1f4dd1c99c0c0b1a133b69c328b/src/crdt/live/live-views.ts#L206-L212

Considered and dismissed by scoring: the missing `isBound` gate on `reEnrollUnsent` is *necessary* (the set only holds actively-edited notes, and recovering a switched-away edit inherently needs a room), the throttled refusal warn still emits ample signal (~20 lines/min/doc on a persistent loop), and `lastRefusedWarnAt` is trivially bounded.

Gates: build clean, full suite (2229 pass / 0 fail incl. random fuzz sim), biome + eslint green.

PR #327, opened 2026-07-25
fix(crdt): end the manifest-validator re-serve stall (hash-aware seq stamp)

Rasbandit:
## Problem

Found via prod Loki on device \`a75644e9\`: the E1 manifest validator was re-serving the same 3-4 notes **~100×/24h**, wedged:

\`\`\`
[client:pull] manifest validator: 3 consumed-but-unrecorded row(s) — rewinding cursor 963 → 883 to re-serve
[client:pull] manifest validator: 4 row(s) still behind after a re-serve — not rewinding again (cursor=963)
\`\`\`

Those notes never converged for the user (stuck showing stale/unsynced) — a *different* bug from the switch-away one in #325.

## Root cause

`validateFromManifest` (#1065) flags a note behind when the manifest seq exceeds the client's recorded seq. When a row advances to a **newer server seq but keeps the SAME `content_hash`** (a meta / seq-only advance), catch-up falls through every branch:
- not **stale** (the seq is newer), and
- not **diverged** (the `content_hash` matches what we recorded).

So the seq is never recorded → the validator re-serves it every poll → forever (cursor bounced 883↔963, `not rewinding again`).

## Fix

Make the validator **content-hash-aware** (same principle as the #296 equal-seq fence): when a behind-by-seq row's manifest `content_hash` equals the recorded `serverHash`, the content is already converged — **stamp the seq and stop re-serving**.

Safety: a shared-seq *fresher* update carries a **different** `content_hash`, so this can never fence out unseen content. Only a genuine hash mismatch (real content divergence) is re-served, exactly as before.

## Test

Differential unit test in `tests/sync-socket-catchup.test.ts`:
- converged row (matching hash, newer seq) → records seq, `behind=0` (RED before this fix: `behind=1`, seq untouched)
- diverged row (hash mismatch) → still `behind=1`, seq untouched

Gates: build clean, **2231 pass / 0 fail** (incl. the #282/#296 fence gates in the sim suite — no data-loss regression), biome + eslint clean.

### DIFF
--- a/main.js
+++ b/main.js
@@ -35,9 +35,9 @@ var __toESM = (mod, isNodeMode, target) => (target = mod != null ? __create(__ge
 )), __toCommonJS = (mod) => __copyProps(__defProp({}, "__esModule", { value: !0 }), mod);
 var __publicField = (obj, key, value) => __defNormalProp(obj, typeof key != "symbol" ? key + "" : key, value);
 
-// node_modules/diff-match-patch/index.js
+// ../../node_modules/diff-match-patch/index.js
 var require_diff_match_patch = __commonJS({
-  "node_modules/diff-match-patch/index.js"(exports, module2) {
+  "../../node_modules/diff-match-patch/index.js"(exports, module2) {
     var diff_match_patch3 = function() {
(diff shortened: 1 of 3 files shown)


Rasbandit:
Folded into #331 (commit cecca9a) so the whole CRDT wave lands as one PR. The fix is unchanged.

PR #328, opened 2026-07-25
chore(crdt): ship convergence lifecycle to Loki via diag()

Rasbandit:
## Why

Prod has **zero visibility** into why a note stages a CRDT convergence but never commits — the deaf-note stall behind both symptoms Todd is hitting: "handshaking so much" churn and mid-session sync wedge. The commit path (`commitCrdtConvergence`) logs its deferral only via `devLog()`, which is console-only and **tree-shaken out of release builds**. Loki sees nothing.

## What

Route the convergence lifecycle to Loki via `rlog().diag()` (backend ships diagnostic-flagged info even at info level — `logs.ex:117`):

- **re-handshake fired** (was `info`) + `attempts` counter — is a handshake firing per file-open?
- **commit DEFERRED** (was `devLog`, invisible) with projected-vs-staged **length + FNV hash only** — never raw note content. Signatures: `projLen ≈ 2×stagedLen` = doubling, `projLen=0` = doc not hydrating, hashes-close = ordering/whitespace.
- **projectedText failure** (was `devLog`)
- **STEP2 committed** (was `info`) + `seq` — does it *ever* commit?
- **channel `crdtJoined` transitions** — `-> false` (socket close + code) / `-> true` (rejoin), to catch the topic wedge.

**Instrumentation only — no logic change.** This gathers the evidence to root-cause the never-commits stall *before* touching the doubling-prone convergence path (the deleted verify-by-text design once caused document doubling). Reference: `relay/Relay/src/client/provider.ts` — in Yjs, applying `syncStep2` **is** convergence; there is no text-verify, which is the direction the eventual fix follows.

## Testing
Build green (tsc + esbuild). Pure log-level changes; no behavioral test applies (asserting "it calls diag" tests a mock). Install the CI pre-release via BRAT, reproduce the wedge, read the `diagnostic` firehose in Loki.

### DIFF
--- a/main.js
+++ b/main.js
@@ -23186,6 +23186,14 @@ var _EngramSyncPlugin = class _EngramSyncPlugin extends import_obsidian25.Plugin
      *  stack. That churn was starving CRDT delivery (empty-flush clobber under a
      *  reconnect that raced note reconciliation). */
     this.liveChannelKey = null;
+    /** Connection identity (backend|account|vault) the PERSISTENT CRDT stack
+     *  (manager + wiring + liveViews + Y.Docs) was built for. Relay model: the
+     *  doc layer outlives the socket. A socket reconnect swaps only the transport
+     *  (a fresh NoteChannel, re-pointed at the surviving wiring via the box) — the
+     *  Y.Docs are NEVER destroyed, so reconnect is a clean syncStep1 diff, not a
+     *  full re-push that doubles the lineage. The stack is torn down ONLY when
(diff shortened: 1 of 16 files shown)


Rasbandit:
Closing: the diagnostics served their purpose (they established the wedge was not the socket, and that client `diag()`/info never reaches Loki - warn does). The lifecycle signal worth keeping is now on `rlog().warn`/`error` in #331; the rest was scaffolding against pre-rewrite code.

PR #330, opened 2026-07-25
fix(crdt): persistent per-note Y.Doc lifetime (bounded LRU) to stop switch-away edit loss

Rasbandit:
## The bug (Todd's repro: switch files, some edits never reach the server)

Rapid file switching in Obsidian silently drops edits. Root cause, caught in prod Loki (device \`a75644e9\`, build \`g894db69\` diagnostics):

\`getDoc()\` re-mints a Y.Doc whenever one isn't resident, while \`onLastViewerRelease\` and \`hibernateIfIdle\` \`closeDoc()\` the doc the instant its editor view closes / a background apply finishes. The channel stays subscribed after you navigate away, so incoming frames (\`startSync\`/\`handleFrame\`/\`hasHistory\`) keep resurrecting docs that hibernation just closed — a **close↔re-mint storm**. The capture showed one open note cycling through 3+ distinct Y.Docs in seconds, ending with the editor bound to the **wrong doc**:

\`\`\`
BIND ok path="New Document Intake.md" docGuid=030ac835 ytLen=8618 edLen=34007
\`\`\`

edLen=34007 = the *previous* note's buffer. After this, sends stop entirely.

## The fix (Relay-style persistent doc)

A note's Y.Doc lives for as long as it fits a bounded LRU working set (\`CrdtManager.residentCap\`, default 64) — **never torn down merely because its editor view closed**. Switching among a handful of files keeps each doc resident, so the editor stays bound to one stable doc.

- **CrdtManager**: LRU recency + protected-set. \`touchLru()\` on every \`entry()\` access evicts the least-recently-used **unprotected, non-in-flight** doc past the cap (persist-safe \`closeDoc\` — IndexedDB store kept). \`protect()\`/\`unprotect()\` pin a live-bound doc.
- **CrdtLiveViews**: \`onBind\` protects; \`onLastViewerRelease\` unprotects instead of \`closeDoc\`.
- **SyncEngine.hibernateIfIdle**: now a no-op (lifetime is the manager's LRU). Heal-room release still resets enrollment; it just no longer eager-closes.

## Tests

TDD: manager LRU tests (resident-within-cap / evict-LRU / never-evict-protected), updated live-views + note-yjs-update + sync-catchup behavior. Full suite green (2232).

Complements #325 (send-layer re-enroll) — that fixed the *send* path; this fixes the *doc-lifetime* layer that was still churning.

### DIFF
--- a/main.js
+++ b/main.js
@@ -16768,6 +16768,20 @@ var _CrdtManager = class _CrdtManager {
      *  it to callbacks. */
     this.docs = /* @__PURE__ */ new Map();
     /**
+     * LRU recency of resident doc IDs — least-recently-used first, most-recent
+     * last. Touched on every `entry()` access (mint or cache hit). Drives
+     * cap-pressure eviction so the resident set stays bounded WITHOUT tearing a
+     * doc down the moment its editor view closes (which caused the switch-away
+     * churn/data-loss). Same key space as `docs`.
+     */
+    this.lru = [];
+    /**
+     * Doc IDs that must never be evicted regardless of LRU position — a live
+     * editor is bound to them. `CrdtLiveViews` calls `protect`/`unprotect` on
+     * bind/release. Same key space as `docs`.
+     */
+    this.protectedIds = /* @__PURE__ */ new Set();
+    /**
      * Per-session set of doc IDs for which at least one inbound server sync
      * frame has been applied (i.e. the STEP2 handshake has completed for the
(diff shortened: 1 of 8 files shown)


Rasbandit:
Superseded by #331. This patches `src/crdt/manager.ts`, which the Relay-model rewrite deletes; the persistent per-note doc lifetime it was adding is now the provider registry's native behaviour (docs stay resident, eviction is sync-gated).

PR #331, opened 2026-07-26
fix(crdt): rebuild client on Relay's model

Rasbandit:
Rebuilds the plugin's CRDT client on Relay's known-good model, and collapses the whole wave into this one PR. Supersedes and closes #327, #328, #329, #330, #332.

## The model

- **One persistent Y.Doc per note.** The provider never tears the doc down on a transport reconnect; it re-advertises with syncStep1 (a state vector), so a reconnect can never re-push full state and fork a lineage. Replaces the bespoke `CrdtManager`/convergence machinery (deleted, along with `CrdtChannel`/`CrdtEnrollment`).
- **Convergence IS `readSyncMessage`.** No text-equality gate, no staged "commit deferred", no seed gate.
- **The live binding is a CM6 `ViewPlugin`** (`src/crdt/live/live-binding.ts`), one instance per `EditorView`, recreated automatically whenever Obsidian rebuilds a leaf's editor. No Compartment for `setViewData` to wipe, no poll, no re-bind race. This structurally removes the file-switch wedge class rather than patching it.

## Edit-safety fixes on top

- **Base-aware 3-way merge in the bind reconcile.** Typing during doc hydration used to forward a whole-text `diff(docText -> editorText)`, deleting any content the editor never saw (a doc hydrated from IndexedDB with server-newer content lost it silently). Now only the typed hunks are patched onto the doc. The LCA is the editor's own text before the user's first keystroke, *not* `BaseStore` - that store is only refreshed on the REST push/pull paths, so it goes stale for CRDT-synced notes and patching a stale base recreates the doubling class. Falls back to the old two-way forward with no base or on any failed hunk.
- **Reading-view edits are captured.** A checkbox toggled in a reading pane calls `previewMode.edit`, which writes the file directly; with an editor pane also open that write is dropped as binding-owned, so the toggle was reverted. Ported from Relay's `ViewHookPlugin` Hook 3, into `CrdtReadingView` (a reading-only leaf has no editor, so no ViewPlugin). Merged, not whole-text-diffed, so a remote edit landing between the pane's render and the click survives.
- **The initial reconcile can no longer stall the binding.** A throw used to skip `goLive()`, leaving `ready` false forever with the drift self-heal unreachable, so the note silently stopped syncing.
- **Manifest-validator re-serve stall** (was #327): a row carrying a newer server seq with an already-recorded content hash is stamped rather than re-served forever.

## Verification

2194 unit tests green, `bun run build`, `biome check`, and `lint:obsidian` all clean. Every commit signed. Pairs with backend #1120.

Follow-up not in scope: awareness/remote cursors (a presence feature with no edit-safety role).


Rasbandit:
### E2E triage (first run) — fix-forward pointer

Plugin-side CI all green (`pr-build`, `build-and-test` unit 2171, CodeQL, lints). `backend/e2e` red is a tight cluster, **all in the idle-note vault-fan-out path** (a note not open in an editor converging via `note_yjs_update` + `applyPushedNoteUpdate`, no enrolled room):

- `test_idle_note_converges_via_fanout_only`
- `test_concurrent_cold_edits_survive_over_fanout`
- `test_cold_send_over_fanout_opens_no_room` — already a known main-flake
- `test_fanout_receive_after_hibernate_rehydrates` — likely asserts the **removed** hibernation contract (`closeDoc`/`hibernateIfIdle` are no-ops in the Relay model now), same shape as the unit tests that had to be retargeted

Everything else e2e-crdt passes (discovery, concurrent edits, delete, live-bound convergence, deaf-note heal, obsidian↔web live, seq-gap-heal) — so the swap is sound for open/live-bound notes; the fix-forward is specifically **idle-note fan-out convergence** under the persistent-doc model.

Rasbandit:
Follow-up: the idle-note fan-out fix landed in cf70aca (decouple room-advertise from transport-connect) + the paired backend e2e test update is **engram-app/Engram#1120** (same branch name → that PR's e2e tests the pair). The cross-triggered `backend/e2e` check *here* runs against backend main, so it stays red on the one retargeted test until #1120 merges; #1120's own e2e is the green gate for the pair.

Rasbandit:
### e2e-crdt GREEN ✅ (17/17)

Paired run [engram-app/Engram#1120 → 30184339681](https://github.com/engram-app/Engram/actions/runs/30184339681): **e2e-crdt: 17 passed, 0 failed**. All four originally-failing idle-fan-out tests pass. Three plugin fixes got there:
1. **Room-free invariant** — decoupled `setAdvertised` (syncStep1/room) from `setConnected` (flush/send), so cold sends + fan-out receives stay room-free (they open a room only when the editor binds them).
2. **Heal-room release** — `reset` now clears the provider's `synced` flag so a re-handshake re-fires `onSynced`→`releaseHealRoom` (the provider fires onSynced only on the first syncStep2).
3. **Create-ack room-free** — `flushHeldState` is a SEND (re-flush the create-gated buffer), not an enroll, so creating a note no longer leaks a permanent room.

**Remaining CI red is pre-existing, not this PR:** `e2e-clerk test_48 receives_update` fails byte-identically on `main` (run 30184376011: 1 failed / 95 passed — a reconnect catch-up of a REST update, engine-agnostic); `headless-protocol` is fail-fast-cancelled on main too. This branch is at parity with main on the non-CRDT jobs and green on the CRDT jobs that exercise the rebuild.

PR #332, opened 2026-07-26
diag: [EDIAG] console breadcrumbs on bind/edit/send path (NOT for merge)

Rasbandit:
Diagnostic-only build so Todd can copy the exact keystroke path from the browser console (plain `console.log` at Default level; `devLog` tree-shakes to a no-op in prod and uses hidden `console.debug`).

Prints `[EDIAG]` lines at each decision point:
- **editor-controller** — `bindTo getYText` (with await ms), `BOUND` / `ABORT stale-epoch` / `ABORT view-switched` / `DEFER unseeded` / `detach` / `deferred onSeed REBIND` / `drift DETACH`
- **note-provider** — `localEdit` (keystroke reached the doc) + `broadcast` (frame sent or buffered)
- **provider-registry** — `providerSend` (gated by create-ack, or sent)

Tells us precisely where a file-switch edit dies: editor never re-binds (no `BOUND`), keystroke never reaches the doc (no `localEdit`), or the frame never sends (`sendOk=false`).

Never merges. Branched off the relay-persistent-stack tip (#331).


Rasbandit:
Folded into #331. The EDIAG scaffolding is stripped (commit 71d29ba) and this branch's work is now on #331's branch as a fast-forward - `lint:obsidian` is clean again now that `ediag.ts` is gone.

PR #334, opened 2026-07-26
feat(settings): collapse 3 diagnostics toggles into one

Rasbandit:
**Parked, not proposed.** Surfaced by a worktree audit — this branch was local-only (never pushed, no PR) and would have been permanently lost on cleanup. Opening as a draft so it's recoverable and discoverable; no claim that it should merge as-is.

## What it does

Collapses three separate diagnostics toggles in plugin settings into a single control. 1 commit, 9 files, +109/-74 — a net simplification of the settings surface, with `tests/settings-migrate.test.ts` (+40) covering migration of existing users off the three old keys.

## Audit notes

- Last touched **2026-07-07** — roughly 66 commits of `main` ago. Expect conflicts and red CI against current `main`.
- The migration test is the load-bearing part: users on the old three-key shape need their settings carried over, so this is not a pure UI change.
- Settings have moved since (CRDT relay rework, enrollment removal), so the three keys this collapses may no longer all exist. **Verify the old keys are still present before rebasing** — if they're gone, the migration path is the only salvageable piece.

## Decision needed

Rebase and verify the three source toggles still exist — or close this and delete the branch, keeping the settings-migration pattern in mind for whenever the diagnostics surface next gets touched.

https://claude.ai/code/session_013fXPYtAKeUBLKjW9QmVRpZ

### DIFF
--- a/main.js
+++ b/main.js
@@ -1410,6 +1410,11 @@ var ApiKeyAuth = class {
 _OAuthAuth.EXPIRY_BUFFER_MS = 6e4;
 var OAuthAuth = _OAuthAuth;
 
+// src/settings-migrate.ts
+function migrateDiagnosticsEnabled(raw) {
+  return raw ? typeof raw.diagnosticsEnabled == "boolean" ? raw.diagnosticsEnabled : !!(raw.remoteLoggingEnabled || raw.diagnosticMode || raw.tracingEnabled) : !1;
+}
+
 // src/error-util.ts
 function errMsg(e) {
   var _a;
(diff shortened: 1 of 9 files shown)


Rasbandit:
Closing — **this already shipped.** Everything this branch introduced is on `main` today, in the same five files it touched:

| Introduced here | On `main` |
|---|---|
| `diagnosticsEnabled` | `src/types.ts`, `src/main.ts`, `src/diagnostics.ts`, `src/tabs/advanced-tab.ts` |
| `migrateDiagnosticsEnabled` | `src/settings-migrate.ts`, `src/main.ts` |
| `src/settings-migrate.ts` | present |
| legacy `remoteLoggingEnabled` / `diagnosticMode` / `tracingEnabled` removed from settings | removed |

`main.ts` also carries the legacy-key drop-on-next-save loop, so the migration half landed too.

### Merging this would be actively harmful
The branch forked at `7e73d95` (#183) and is **114 commits behind** `main`. Its diff against `main` is **199 files, +25,775 / −47,892** — that is not "what it adds", it's the accumulated divergence. Merging or naively rebasing it would revert a very large amount of shipped work, including the CRDT Relay rewrite.

The branch is left in place — nothing is lost, and the surviving artifact is the shipped implementation on `main`, not this one.

_Context: surfaced by the same 2026-07-26 worktree-cleanup sweep that parked #333. #333 was genuinely unmerged and has now shipped; this one had already been overtaken._

PR #336, opened 2026-07-26
fix(auth): never send a request with an empty bearer token

Rasbandit:
## Why

`engram-prod-loki-auth-failure-burst` has fired 14 times in the last 7 days. It is not credential stuffing and not the socket-signature class the existing runbook covers — it is our own plugin, unlinked, looping 401s.

**Chain:** an install with no auth provider (nulled by `clearAuthAndPromptRelink`, or by a backend switch) and an empty `apiKey` fell through to `getAuthToken() → this.apiKey === ""`, so `sendRequest` built `Authorization: Bearer ` with nothing after it. HTTP strips trailing whitespace from field values, so the backend sees a bare `Bearer`, misses the `["Bearer " <> token]` match in `EngramWeb.Plugs.Auth.authenticate/1` and logs `reason=no_auth` — one warn line in `category=auth` per request.

Verified against prod:

| probe | logged reason |
|---|---|
| `Authorization: Bearer ` (empty) | `no_auth` |
| `Authorization: Bearer abc` | `signature_error` |

Nothing gated sync, remote logging or the trace beacon on auth state (`grep isAuthenticated() src/main.ts src/sync.ts` → no guards), so such an install never stops. Prod, 24h to 2026-07-26: **890 `no_auth` 401s** from three installs — `POST /api/folders` ×648, `/api/logs` ×102, `/api/telemetry/spans` ×98, `/api/attachments` ×14. Note the middle two: the client's log push and beacon carry its own complaints about the 401s, so the noise feeds itself.

## What changed

- `getAuthToken()` throws instead of returning an empty credential. One guard on the shared request path covers every caller — folders, notes, attachments, logs, search — rather than gating each one.
- The beacon transport thunk drops its batch while `lastToken` is empty. It posts with `window.fetch`, so it bypasses the guard above and would 401 on its own.

No version bump — release-please owns that.

## Tests

3 new cases in `tests/api.test.ts`: empty apiKey and empty provider token both reject **without touching `requestUrl`**, and an unauthenticated beacon flush issues no `fetch`.

Full suite 2197 pass / 1 skip / 0 fail, `bun run build`, `biome ci`, `lint:obsidian`, `lint:css` all clean.

> One caveat, stated plainly: the very first (cold) run of the suite in this worktree reported 1 failure whose identity I did not capture, and 8 consecutive runs since have been green — including replays of both `sim/random` seeds from that run, which pass. Unresolved rather than dismissed; CI is the next look.

## Follow-ups not in scope

The alert rule is deliberately untouched — volume should fall to ~0 once this ships, and retuning now would mask the client bug. Separately, the `[client:auth] OAuth refresh failed` timeouts / `ERR_NAME_NOT_RESOLVED` / `ERR_QUIC_PROTOCOL_ERROR` lines in prod are a different (network-side) story.

https://claude.ai/code/session_013fXPYtAKeUBLKjW9QmVRpZ

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -172,6 +172,7 @@ Required status checks on `main`: `build-and-test`, `version-check / version-che
 - V8 OOM prevention on large-vault operations → `docs/context/v8-oom-prevention.md`
 - Editor-binding stale-buffer race (note content copied into a DIFFERENT file on file-switch, PR #194 — bindTo await gap, sync-detach-before-await + bindEpoch + drift view-identity-guard fix) → `docs/context/crdt-editor-bind-race-pollution.md`
 - Missed CRDT delivery healing (catch-up convergence: id adoption parity, base_hash CAS 409, pull backfill, socket vault-catchup via `catchupViaSocket()`) → `docs/context/sync-catchup-convergence.md`
+- Prod `auth-failure-burst` alert traced to our own client (`Bearer ` with an empty token logs `reason=no_auth`, not `signature_error`; unlinked installs loop 401s and the log push re-reports them) → `docs/context/empty-bearer-no-auth-401-loop.md`
 - Convergence sim tier fidelity gaps (`tests/sim/` — differential gate pays rent; seeded random-op suite does NOT converge, kept as a runnable tool not a test) → `docs/context/crdt-convergence-sim-fidelity-gaps.md`
 
 @/home/open-claw/documents/code-projects/ops-agent/docs/self-updating-docs.md
--- a/docs/context/empty-bearer-no-auth-401-loop.md
+++ b/docs/context/empty-bearer-no-auth-401-loop.md
@@ -1,0 +1,77 @@
+# Context Doc: `Bearer ` with no token → `reason=no_auth` 401 loops
+
+_Last verified: 2026-07-26_
+
+## Status
+**Fixed** — plugin PR #336 (`fix/no-empty-bearer-requests`). Alert rule deliberately unchanged.
+
+## Symptom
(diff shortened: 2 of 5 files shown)


Rasbandit:
## Merging with `backend/e2e` red — evidence that the red is not this diff

Run [30224045003](https://github.com/engram-app/Engram/actions/runs/30224045003) (after #335 landed): `1 failed, 95 passed` — `test_82_resumed_device_stale_idmap.py::test_resumed_device_with_stale_idmap_syncs_both_ways`, `TimeoutError: File E2E/Resumed-pull-seed.md did not contain 'original' within 120.0s`. test_48 (the earlier failure here) now **passes** — #335 fixed it.

This change cannot be the cause, from the run's own `ci-debug-` artifacts:

- `Not authenticated: no API key or access token` — the only new error this PR introduces — appears **0 times** in `docker-compose.log` and `pytest-e2e.log`. The guard never fired.
- **0** 401s / `Unauthorized` in the run. The empty-bearer path this PR closes was never exercised.
- The failing seed note produces no client log line at all, matching the shape documented in engram#1130.

And e2e-clerk is rotating a single delivery-timeout failure per run on `main` independently of this branch. Backend `main`'s own push run [30218917759](https://github.com/engram-app/Engram/actions/runs/30218917759), same sha, failed **two**: `test_34_folder_rename_propagation::test_folder_rename_new_paths` (`received=yes materialized=no`) and test_48. That is engram#355.

Unit side is green: 2197 pass / 1 skip / 0 fail locally across 9 runs plus CI `build-and-test`.

Merging on owner decision with the flake left tracked in engram#355, not silenced here.

PR #349, opened 2026-07-29
fix(logging): remoteLogLevel=debug actually reaches Loki

Rasbandit:
Setting `remoteLogLevel` to `debug` made the plugin **send** info-level entries that the backend then silently dropped. Verbose logging produced nothing to read.

## Why

The backend deliberately keeps client logs out of Loki below `warn` — `Engram.Logger.Category` omits `:client` from `@info_to_loki` so a whole plugin fleet can't flood it:

```elixir
# :client is NOT here: re-emitted plugin info would flood Loki; verbose client
# entries opt in per-entry via a loki_ship override (see Logs.insert_logs).
@info_to_loki [:billing, :crypto, :lifecycle, :oban, :boot, :websocket]
```

It offers exactly one escape hatch, in `Logs.reemit_to_logger`:

```elixir
if entry["diagnostic"] == true do
  Keyword.put(meta, :loki_ship, true)
```

`RemoteLogger` only ever set `diagnostic` from `diag()`, never from `info()`. So the setting and the escape hatch never met.

## Cost

Two investigations this session. The code path that recreated a deleted note (`onLastViewerRelease` → `flushToDisk` → `flushFromCrdt`) logs at info and left **no trace at all** — we had to ship a temporary warn-level tracer to find it, then strip it again.

## Fix

Sub-`warn` entries carry `diagnostic: true` when the threshold is `debug`.

- Gated on `debug` specifically, **never** the `info` default — the flood guard still holds for anyone who hasn't asked for the firehose.
- `warn`/`error` already ship via `loki_ship?`, so they stay unflagged rather than padding the payload.

No backend change: this uses machinery that already existed.

## Tests

RED-verified. `debug` flags info; the default does not; `warn` stays unflagged even at `debug`.

Suite **2213 pass / 0 fail**; build + biome + eslint green.

https://claude.ai/code/session_01N62e5eq31DoffeCH2LJDFM

### DIFF
--- a/main.js
+++ b/main.js
@@ -942,7 +942,7 @@ var LEVEL_SEVERITY = {
       platform: this.platform,
       seq: this.seq++
     };
-    stack && (entry.stack = stack), this.connId && (entry.conn_id = this.connId), this.deviceId && (entry.device_id = this.deviceId), this.vaultId && (entry.vault_id = this.vaultId), diagnostic && (entry.diagnostic = !0), this.buffer.push(entry), this.buffer.length > MAX_BUFFER && this.buffer.splice(0, this.buffer.length - MAX_BUFFER), this.buffer.length >= FLUSH_THRESHOLD && this.flush();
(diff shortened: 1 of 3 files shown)


Rasbandit:
Consolidated into #348 — one PR per initiative. The commit (`c62e1a0`) is merged into that branch unchanged; nothing is lost. Closing to keep a single plugin PR to review and land.

Issue #360, opened 2026-07-31
Ops-hygiene bundle from Relay: HasLogging, flag schema, Patcher registry, promise tracker, local log file


Rasbandit:
From the Relay audit (`docs/context/relay-pattern-audit.md` in engram-workspace). Relay finding **H**.

Six small independent items. Each is a standalone PR; take them opportunistically when touching nearby code.

### 1. `HasLogging` base class
`Relay/src/debug.ts` — every class extends it and gets `this.debug/log/warn/error` pre-tagged with its own
name, retagged on rename via `setLoggers(context)`. Ours: manual `rlog().info("channel", …)` string tags at
every call site, which drift.

### 2. Feature-flag schema with categories
`Relay/src/flags.ts` — flags carry `{ default, category: "labs" | "debugging" | "danger", title,
description }`. Danger flags render in a marked zone and are hidden unless debugging is on. Adding a flag
without a schema entry is a **type error**. `flagManager.ts` also accepts **server-pushed overrides**.
Ours: plain booleans in settings, no gating, no server control.

### 3. `Patcher` singleton
`Relay/src/Patcher.ts` — one registry for every `monkey-around` patch, refuses duplicate patches of the same
method on the same target, and warns at startup if patches survived a previous unload. Ours: direct patching
in `src/crdt/live/obsidian-internals.ts` with no registry.

### 4. `trackPromise` leak tracker
`Relay/src/trackPromise.ts` — labelled in-flight promise registry plus a 1,000-entry ring of completions,
**compiled out entirely unless `BUILD_TYPE === "debug"`**. Answers "what is still pending" at unload.
Pairs with #355.

### 5. Local rotating log file
`Relay/src/debug.ts` — ring buffer → batched append → size-capped rotating log file in the vault, with
`allowStackTraces` off by default to avoid retention leaks. We are remote-only (Loki), so an offline user
has nothing to send us.

### 6. Error taxonomy
- `Relay/src/S3Error.ts` — typed error with `retryable` derived from S3 codes + HTTP 408/429/5xx;
  `isRetryableS3Error` gates requeue. Ours: ad-hoc status checks scattered through `api.ts`/`sync.ts`.
- `Relay/src/UserFacingError.ts` — recursively unwraps nested / JSON-stringified errors into one sane user
  string, cycle-safe, 300-char cap. Ours: `src/error-util.ts`, 10 lines.

### Already ported, no action
[...]

Issue #359, opened 2026-07-31
Unify the sync queue: priorities, cancellation, retry policy, and a 'why is this queued' reason


Rasbandit:
From the Relay audit (`docs/context/relay-pattern-audit.md` in engram-workspace). Relay finding **F**.

## Problem

Our queueing is split three ways and the pieces don't talk:

- `src/offline-queue.ts` — dedup `Map`, oldest-first, no priority, no cancellation
- `src/issue-store.ts` — terminal failures, deliberately kept out of the retry loop (this part is good and
  should survive)
- `src/sync.ts` — `activePushCount` / `maxConcurrentPushes` / `pushWaiters` / `releasePushSlot`, with
  progress surfaced via toasts (`flushFailureSummaryToast`, `flushAttachmentLimitedToast`,
  `flushDegradedNotice`)

## What Relay has

`Relay/src/BackgroundSync.ts` — one queue with `QueueItem`, `SyncGroup`, `GroupProgress`,
`BackgroundSyncFailure`, plus:

- `cancelDocumentWork(guid)` — pulls a doc out of every queue and resolves its waiters
- `requeueRetryableSync` bounded by `MAX_PROVIDER_SYNC_RETRIES`
- `getFolderSyncSnapshot` / `getFolderPillProgress` for the status UI
- **`queuedReasonForSnapshot`** — answers *"why is this still queued"* with `not connected` instead of
  leaving the user watching a spinner

And `MergeManager`'s wake queue: a `WakePriority` enum, bounded concurrency, and priority *upgrade* for a
guid already in the queue.

## Proposal

1. One queue owning both retryable network work and CRDT room work, with a priority enum
   (open-document > user-initiated > background catch-up).
2. `cancel(noteId)` so a delete/rename drops queued work instead of racing it.
3. Retry policy in one place, with `IssueStore` still the terminal sink (no change to its role).
4. `queuedReason()` surfaced in the Sync Center — cheapest item here, removes a support-ticket class.
5. Replace the three ad-hoc toast flushers with queue-derived progress.

Relates to engram-workspace#10 (Sync UX: progress bar, summary, retry for bulk sync).

Refs: #99

Rasbandit:
Part of the Relay-pattern migration epic: engram-app/engram-workspace#142

Issue #358, opened 2026-07-31
Extract a SyncedFile object — retire the ~19 path-keyed side tables on SyncEngine


Rasbandit:
From the Relay audit (`docs/context/relay-pattern-audit.md` in engram-workspace). Relay finding **B**.

## Problem

`src/sync.ts` is **7,517 lines** with a single `SyncEngine` class carrying **104 fields**, ~19 of them
`Map`/`Set` containers keyed by path or note_id:

`pushing`, `recentlyPushed`, `remotelyDeleted`, `recentlyFlushed`, `recentlyDeleted`, `pulling`,
`debounceTimers`, `confirmedNoteIds`, `crdtRehandshakeAttempts`, `pendingConvergence`, `crdtHealCooldown`,
`crdtHealTrailingTimers`, `pendingQueueDeliveries`, `manifestPathOwners`, …

Every one is a per-file field hoisted into a global side table. Consequences:

- no single place to ask "what is the state of this file"
- TTL bookkeeping (`markWithTtl`) substituting for ownership
- cleanup must remember to touch all 19 maps — which is why `Lifetime` had to be introduced at the
  registry layer at all

## What Relay has

`Relay/src/IFile.ts` — 15 lines:

```ts
export interface IFile {
  guid: string; path: string;
  move(newPath, sharedFolder): void;
  connect(): void; disconnect(): void;
  cleanup(): void; destroy(): void;
}
```

implemented by `Document`, `SyncFile`, `SyncFolder`. Each owns its own provider, its own `Lifetime`, its
own persistence, its own merge state.

## Proposal

Introduce `SyncedFile` (path, noteId, kind, `Lifetime`, and the per-file state). Migrate the side tables
onto it **one map per PR** — this is mechanical and independently shippable, no big-bang refactor.

Suggested order (least entangled first): `debounceTimers` → `pushing`/`pulling` → the three `recently*`
TTL maps → `crdtRehandshakeAttempts`/`crdtHealCooldown`/`crdtHealTrailingTimers` → `confirmedNoteIds`
(coordinate with #305) → `pendingConvergence`.

Blocked-ish on nothing, but lands more cleanly after #357 (LCA gives `SyncedFile` its most important
field).

Refs: #305, #357

Rasbandit:
Part of the Relay-pattern migration epic: engram-app/engram-workspace#142

Issue #357, opened 2026-07-31
Persist an LCA per note and make applyLocalEdit a real 3-way merge


Rasbandit:
From the Relay audit (`docs/context/relay-pattern-audit.md` in engram-workspace). Relay finding **C** (partial port).

## Problem

`applyLocalEdit` diffs disk content against the **current doc state**, because we have no recorded base.
That is why `src/crdt/provider-registry.ts:283-309` needs a 3-attempt retry loop that ticks `remoteSeq`
and awaits `pendingFlush` just to decide whether a remote merge interleaved the disk read:

```ts
if (reread) {
  let stable = false;
  for (let attempt = 0; attempt < 3 && !stable; attempt++) { ... }
  if (!stable) return null; // live remote storm — skip the stale diff
}
```

That loop is a workaround for a missing base, and it fails open (returns `null`, dropping the edit) under
load.

## What Relay has

`Relay/src/merge-hsm/persistence/HSMStore.ts` persists, per document:

```ts
{ guid, path,
  lca:  { contents, hash, mtime, snapshot },   // last common ancestor
  disk: { hash, mtime },
  localSnapshot, lastStatePath, deferredConflict,
  fork: { base, localSnapshot, remoteSnapshot } }
```

With `lca.contents` in hand the merge is a genuine 3-way (they use `node-diff3`), so "did a remote merge
land during my disk read" stops being a question that needs answering.

## Proposal

1. Extend our existing per-path sync state (`crdtHead`, `recordCrdtBaseline`, `isUnchangedSynced`) into a
   full LCA record: `{ contents, hash, mtime, snapshot }`, persisted alongside `crdtHead`.
2. Update the LCA at exactly one point — a completed remote merge flush (`onFlushToDisk` success).
3. Rewrite `applyLocalEdit` as `diff3(disk, lca.contents, projectedText)`.
4. Delete the stale-snapshot retry loop, `remoteSeq`, and the `pendingFlush` await in that path.

## Also worth taking: name the states

Relay's `machine-definition.ts` names 22 states. Four of them are literally our bug titles:
`idle.diverged`, `idle.localAhead`, `idle.diskAhead`, `idle.recoverLCA`. We should adopt the **enum**
(a `SyncState` union per note, surfaced in the debug snapshot) without porting the 5,945-line HSM.

## Explicitly out of scope

`MergeHSM.ts` itself, the interpreter, the visualizer, `undo/OpCapture`. More machine than a single-user
multi-device product needs.

[...]

PR #361, opened 2026-07-31
feat: Relay-pattern migration — six patterns adopted from the Relay audit

Rasbandit:
Implements the plugin-side work from the Relay audit (engram-app/engram-workspace#142,
`docs/context/relay-pattern-audit.md`). One commit per issue, ordered so each is
independently reviewable.

| Commit | Issue | What |
|---|---|---|
| 1 | #360 | ops-hygiene primitives: feature-flag schema, HasLogging, trackPromise |
| 2 | #355 | debug snapshot API — every copy of a note side by side |
| 3 | #356 | record + replay the CRDT sync seams |
| 4 | #359 | queue priority, synchronous cancel, queued-reason |
| 5 | #357 | merge disk edits against the LCA instead of current doc state |
| 6 | #358 | per-file sync state as an object (first slice) |

`2367 pass, 0 fail` · tsc, eslint, stylelint, biome all clean.

## Things worth a reviewer's attention

**The LCA store already existed.** `src/base-store.ts` persists the last-synced
content of each note and its own docstring calls it "the common ancestor for 3-way
merge conflict resolution", with LRU eviction and a 50MB cap. Nothing in the CRDT
path was reading it. So #357 turned out to be a merge problem, not a persistence
problem: `seedContentInto` ends in a TWO-way `diffIntoYText(text, diskBody)` that
forces the Y.Text to equal the disk snapshot, deleting any remote op that landed
between the file read and the diff. That is the stale-snapshot revert class, and
the reason `applyLocalEdit` carries a 3-attempt retry loop. With a base we apply
the disk's delta *relative to the base* instead, and remote work survives by
construction. `diff-match-patch` was already a dependency.

**The queue was not rewritten.** Relay's BackgroundSync is 2450 lines, but we
already had dedup, a persisted retry cap, oldest-first ordering, and IssueStore as
a terminal sink outside the retry loop. Only three things were missing and only
those were built.

**Three items from #360 were deliberately skipped**, with reasons in the commit:
the Patcher registry (we have 2 patch sites, both already returning unpatchers),
the local rotating log file (we have Loki + a dev ring buffer), and the error
taxonomy (deferred to where it has a caller).

## Flags

Both new flags are **off by default** and only visible with diagnostics enabled.
`crdtLcaMerge` is `danger` category — it changes how concurrent edits resolve and
wants a real soak before it becomes the default. The two-way path stays fully
reachable, and a test pins its lossy behaviour on purpose so it cannot rot while
it is still what ships.

## Not done here

Retiring the `applyLocalEdit` retry loop — it is still the path whenever
`crdtLcaMerge` is off, so deleting it now would leave the default configuration
unguarded. And 15 of the 19 side tables in #358; `debounceTimers` in particular
has rename re-arming logic that deserves its own change.

Backend decision engram-app/Engram#1146 (vault index as a CRDT) is independent of
all of this and still open.

https://claude.ai/code/session_01N62e5eq31DoffeCH2LJDFM

### DIFF
--- a/src/content-hash.ts
+++ b/src/content-hash.ts
@@ -1,0 +1,15 @@
+/** Fast string hash (FNV-1a 32-bit). Not cryptographic — just for content
+ *  change detection.
+ *
+ *  Lives in its own module rather than in sync.ts so a consumer that needs only
+ *  a content hash (the debug snapshot, and shortly the LCA record) does not pull
+ *  the entire sync engine into its import graph. `sync.ts` re-exports it, so
+ *  existing importers are unaffected. */
+export function fnv1a(s: string): number {
+	let h = 0x811c9dc5;
+	for (let i = 0; i < s.length; i++) {
+		h ^= s.charCodeAt(i);
+		h = Math.imul(h, 0x01000193);
+	}
+	return h >>> 0;
+}
--- a/src/crdt/lca-merge.ts
+++ b/src/crdt/lca-merge.ts
@@ -1,0 +1,55 @@
+/**
+ * Apply a disk edit to a Y.Doc's text WITHOUT reverting remote work.
+ *
+ * The problem this replaces (#357). `seedContentInto` ends in
+ * `diffIntoYText(text, diskBody)` — a TWO-way diff that forces the Y.Text to
+ * equal the disk snapshot. Any remote op that landed between reading the file
+ * and running the diff is absent from that snapshot, so the diff deletes it.
+ * That is the stale-snapshot revert class (e2e test_83), and it is why
+ * `applyLocalEdit` carries a 3-attempt retry loop that ticks `remoteSeq` and
+ * awaits `pendingFlush` just to detect the race — a workaround for a missing
+ * merge base.
+ *
+ * With a base (the last-synced content, which `BaseStore` already persists —
+ * its own docstring calls it "the common ancestor for 3-way merge") the right
+ * operation is not "make the doc equal disk" but "apply the disk's delta
+ * RELATIVE TO THE BASE onto the doc". Remote ops survive because they exist in
+ * the doc and are untouched by that delta.
+ *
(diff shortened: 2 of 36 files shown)


Rasbandit:
### Code review

Self-review of this branch (author-reviewed, so treat as a first pass rather than independent eyes). Found 2 real bugs, both fixed in 4d119a6.

1. `SyncedFileTable` — a stale expiry timer could delete an unrelated newer entry. The closure captures the path it was armed for, but `rename()` moves the same object to a different key. Arm a marker, rename the file, then touch the OLD path again (a new file created at the freed name): the pre-rename timer fires and deletes the NEW entry, silently dropping its echo suppression. Now identity-checked before deleting.

https://github.com/engram-app/Engram-obsidian/blob/aa96829e1a778f9867add68999318eacf43e24ff/src/synced-file.ts#L77-L86

2. The LCA branch bypassed the adopt-first gate. `mergeDiskOntoDoc` short-circuits `base === disk` to the doc's own projection, which for a history-less doc is empty — and the caller records `fnv1a(returned)` as the last-synced baseline (`applyCrdtCreateAck`). A fresh note whose disk matched its base therefore recorded the hash of `""`, making `isUnchangedSynced` false from then on and re-opening the #846 divergent-lineage doubling path. Fixed by gating the LCA branch on the doc actually having history — a history-less doc holds no remote work for the merge to protect, so adopt-first plus `seedOnce` is already correct for it.

https://github.com/engram-app/Engram-obsidian/blob/9374c159639a5b90d92c3cf1fe36c649e67472c9/src/crdt/provider-registry.ts#L305-L312

Worth noting: the first regression test for (1) passed for the wrong reason — `fireAll()` also expired the marker under assertion. Added `fireOldest()` so it fires only the pre-rename timer.

Suggest running `/code-review ultra 361` for an independent pass before merge, given the author-review caveat.

PR #379, opened 2026-08-03
fix(settings): isConnected guard on the progress-bar render

### LINKED ISSUE #375: fix(settings): isConnected guard on the progress-bar render (1.13+ teardown-without-hide)
Pre-existing issue surfaced by review pass 2 on PR #374 (scored out of that PR as pre-existing).

On Obsidian 1.13+, the settings UI renders via `getSettingDefinitions()` into a row Obsidian can tear down and recreate WITHOUT calling the tab's `hide()` — and `hide()` is where `uninstallProgressBar()` runs. The installed `onSyncProgress` wrapper then keeps rendering progress into detached DOM nodes (and retains them) until the next settings open reinstalls it.

`renderStatus()` already guards with `statusEl?.isConnected` for exactly this teardown-without-hide path (see the `rerender()` doc comment in `src/settings.ts`). The progress render closure needs the same guard: check `progressContainer.isConnected` at the top of the render callback (the wrapper should still forward to `prev` regardless).

Smallest of the four review follow-ups; a good first PR of the wave.

Rasbandit:
## Summary

On Obsidian 1.13+ the settings UI renders via `getSettingDefinitions()` into a row Obsidian can tear down and recreate WITHOUT calling the tab's `hide()`, so `uninstallProgressBar()` never runs and the installed `onSyncProgress` render closure keeps rendering into (and retaining) detached DOM nodes until the next settings open.

Fix: guard the render callback with `progressContainer.isConnected`, exactly like `renderStatus()` already does for the same teardown-without-hide path. The `installProgressBar` wrapper still forwards to `prev` regardless (chain behavior untouched, covered by the existing `settings-progress-chain` tests).

To make the guard testable, the render closure is extracted into a DOM-agnostic `makeProgressBarRender` factory (same pattern as `renderCompletionSummary`); body otherwise unchanged.

## Tests

- New `tests/settings-progress-render.test.ts`: detached no-op (including `complete`), reattach recovers, connected render, prevTotals cleared after complete.
- Full suite 2394 pass; build + biome/eslint/stylelint clean.

Fixes #375

https://claude.ai/code/session_01GjwVE1c1F691yNPKtkjkpJ

### DIFF
--- a/main.js
+++ b/main.js
@@ -18460,6 +18460,28 @@ function renderSyncCenterTab(ctx) {
 }
 
 // src/settings.ts
+function makeProgressBarRender(progressContainer, progressLabel, progressBarInner, getPlannedPhases) {
+  let prevTotals = /* @__PURE__ */ new Map();
+  return (progress) => {
+    var _a, _b, _c;
+    if (!progressContainer.isConnected) return;
+    if (progress.phase === "complete") {
+      progressContainer.removeClass("is-active"), prevTotals.clear();
+      return;
+    }
+    progressContainer.hasClass("is-active") || prevTotals.clear(), progressContainer.addClass("is-active");
+    let planned = (_a = getPlannedPhases()) == null ? void 0 : _a.find((p) => p.phase === progress.phase), { current, total, pct } = settingsBarCounts(
+      progress,
+      planned,
+      (_b = prevTotals.get(progress.phase)) != null ? _b : 0
+    );
+    prevTotals.set(progress.phase, total);
+    let phaseLabel = (_c = PHASE_FALLBACK_LABEL[progress.phase]) != null ? _c : progress.phase, failedSuffix = progress.failed > 0 ? ` (${progress.failed} failed)` : "";
+    progressLabel.setText(
+      total > 0 ? `${phaseLabel}... ${current}/${total}${failedSuffix}` : `${phaseLabel}... ${current}${failedSuffix}`
+    ), progressBarInner.style.width = `${pct}%`;
+  };
+}
 var EngramSyncSettingTab = class extends import_obsidian23.PluginSettingTab {
   constructor(app, plugin) {
     super(app, plugin);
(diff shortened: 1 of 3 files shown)


Rasbandit:
### Code review

No blocking issues. One observation, and one thing that needed fixing before merge.

**Needed fixing (done in ce4f101):** the branch was 2 commits behind after #380 and #382 merged, and its committed `main.js` contained zero occurrences of `adoptCreateAck` while `main`'s had five. The green checks were therefore run against a pre-merge base. Merged `main` in and rebuilt: the bundle now carries both `adoptCreateAck` (5) and `makeProgressBarRender` (2), and the rebuild produced a byte-identical artifact to the textual merge, so no corruption. Full suite 2415 pass, biome/eslint/stylelint clean.

**Observation, not blocking:** the `isConnected` guard now also short-circuits the `complete` branch, so a completion arriving while detached no longer clears `is-active` or `prevTotals`. If the *same* container node were ever reattached, `is-active` would still be set, the `if (!hasClass("is-active")) prevTotals.clear()` guard would not fire, and a stale denominator could leak into the next sync's fallback rows.

In practice this looks unreachable: the precedent this PR cites establishes that a detached container is discarded and re-rendered fresh on next open, not reattached.

https://github.com/engram-app/Engram-obsidian/blob/ce4f101c8a1cbd2a04a9b56ea0e6b0bd0a9e0ca9/src/settings.ts#L146-L153

The mild inconsistency is that `tests/settings-progress-render.test.ts` pins reattach-recovery as intended behavior ("reattached container renders again on the next event") while the adjacent test comments assume the fresh-container model ("next open re-renders fresh"). Worth deciding which model the tests mean to encode, but the shipped behavior matches the established pattern either way.

Verified `activeSyncPhases` is declared `PlannedPhase[] | null`, matching the new `getPlannedPhases` parameter exactly — no undefined-vs-null gap. The extraction is otherwise behavior-preserving.

Issue #376, opened 2026-08-03
refactor: split sync.ts into 7 modules (prereq: syncState mutators + CrdtPorts)


Rasbandit:
Follow-up from the 2026-08-01 repo review (remediated in PR #374; this is the one large item deliberately deferred there).

`SyncEngine` (`src/sync.ts`, ~7,400 lines) is a god class: ~45 mutable fields and 15 `setX` late-injection setters. Split along the seams the review mapped:

- `push.ts` (~1,300): `handleModify/Delete/Rename`, `pushFile`, `pushGenesisBatch`, `partitionGenesis`, `pushPartitioned`, plan pre-gates + batch toasts
- `apply.ts` (~1,400): `handleStreamEvent`, `applyOp`, `applySyncChange`, `applyChange`, `applyAttachmentChange`, relocation + materialize, `eventToOp`
- `catchup.ts` (~700): `catchUp`, `catchupViaSeqReplay(+Exclusive)`, `runSeqReplayOnce`, `enumerateServerState`, manifest validate/reconcile, cursors
- `converge.ts` (~700): `socketConverge`, `fireCrdtReHandshake`, `commitCrdtConvergence`, heal machinery, convergence maps
- `queue-flush.ts` (~450): `flushQueue`, `drainUntilStable`, terminal issues, offline state machine
- `vault-fs.ts` (~250): folder/file creation, `removeEmptyFolders`, drift-conflict copies
- `sync-util.ts`: already-pure module helpers

**Prerequisites (valuable standalone, do first):**
1. Centralize the 30+ `syncState` write sites behind 3-4 named mutators (`recordPushed`, `recordConverged`, `seedCasBase`, `dropPath`) — the sites use subtly different merge semantics today.
2. Collapse the 15 CRDT setters into one `CrdtPorts` object passed once. Keep the setters as thin shims: the backend e2e harness and `tests/sim/replica.ts` call them.

Needs supervised work with e2e coverage per phase; do NOT attempt as one mechanical pass.

PR #380, opened 2026-08-03
refactor(sync): the two prerequisites for the sync.ts split

Rasbandit:
Both prerequisites listed in #376, which the issue calls out as valuable standalone. The split itself (7 modules) is **not** in this PR.

## 1. Centralize `syncState` writes behind named mutators

The ~30 raw `syncState` set/delete sites used subtly different merge semantics — full replace vs merge-at-stamp-time vs merge-onto-a-stale-snapshot — with nothing at the call site saying which. Every write now routes through four named mutators, so each site's choice is explicit and greppable:

| Mutator | Semantics |
|---|---|
| `stampSyncedRow` | REPLACE the row (lineage (re)known wholesale) |
| `patchSyncedRow` | MERGE onto the row read at stamp time |
| `recordCrdtBaseline` (+`markCreated`) | echo-baseline stamp, genesis `crdtHead` sentinel flip folded in |
| `dropPath` (+`dropBase`) | row drop with explicit CAS-base pairing |

Behavior-preserving **except one deliberate tightening**, called out inline: `pushFile`'s CRDT-op branch merged its echo-baseline onto a PRE-AWAIT `syncState` snapshot, silently dropping any `version`/`serverHash` a concurrent converged stamp wrote during the awaited `routeModify`. It now merges at stamp time like every other site.

`wipePerVaultState` keeps its raw `clear()` — it is itself the named wipe mutator.

## 2. Collapse the 15 CRDT setters into `setCrdtPorts`

`main.ts` wired the CRDT stack through 15 late-injection setters spread across three lifecycle stages (boot, channel-join, teardown), so "what is wired, and when" was unreadable at the call site.

One `setCrdtPorts(patch)` primitive now takes each stage's subset:
- only keys **present** in the patch are assigned, so a later stage can never clobber an earlier stage's ports
- an explicit `null` clears a port (the teardown semantic)
- `main.ts` drops from 15 scattered calls to 7 stage-shaped patches

**The legacy `setX` methods stay as thin shims.** They are harness-facing surface, driven directly by `engram/e2e/headless/run.ts`, `tests/sim/replica.ts`, and ~30 unit test files. Only the call sites collapse.

Also hoists the two inline structural types that were spelled out twice each (`CrdtCreateBatchFn`, `CrdtCatchupSinceFn`).

## Tests

New `tests/crdt-ports.test.ts` covers the patch semantics: a patch assigns only the keys it names, a later stage does not clobber an earlier one, an explicit null clears, and each harness-facing shim still lands in the same port.

`tests/main-catchup-wiring.test.ts`'s hand-rolled `SyncEngine` double listed the 8 setters `connectChannel` calls; it now accepts the one port patch. Those stub setters were all no-ops, so this is a surface mirror, not a loosened assertion.

## Verification

- `bun run build` (tsc + esbuild) clean
- `bun test` — 2402 pass, 1 skip, 0 fail
- `biome ci src tests`, `lint:obsidian`, `lint:css`, `lint:lockfile` all clean

Refs #376


Rasbandit:
### Code review

Found 1 issue, fixed in 294d479:

1. The `CrdtPorts` doc states a universal contract its own interface breaks: "A key must be PRESENT in the patch to be assigned; an explicitly-null key clears that port." That held for 14 of the 15 ports. `liveBound` was typed without `| null` and assigned on a truthiness check (`if (ports.liveBound)`) rather than a presence check, so it could not be cleared at all and the documented rule was simply false for it.

https://github.com/engram-app/Engram-obsidian/blob/8f37947fe6e87f4119df33f790e57df4bd5d1842/src/sync.ts#L342-L344

https://github.com/engram-app/Engram-obsidian/blob/8f37947fe6e87f4119df33f790e57df4bd5d1842/src/sync.ts#L477-L479

Fixed by making the port uniform rather than documenting the exception, so the interface has one rule with no carve-outs. Nulling it restores the `() => false` default instead of leaving the port empty, since `isLiveBound` is called unconditionally on the push path. No prod behavior change: `main.ts` only ever sets this port, and its closure already self-nulls to `false` through optional chaining once `crdtLiveViews` is torn down. Pinned by a new test.

Checked and found clean: `syncState` mutator semantics at every converted call site (replace vs merge-at-stamp-time vs drop, including both `dropBase: false` sites), the 15 collapsed wiring calls against each setter's originating fix commit (#915, #970, #191, #201), and prior review comments on the CRDT create-ack lineage.

Rasbandit:
### Code review

Follow-up pass over `294d479` (the fix from the previous review, which was itself unreviewed).

No issues found.

Verified specifically: changing `if (ports.liveBound)` to `if ("liveBound" in ports)` introduces an `in`-vs-truthiness edge case, since `in` is true for a key present with value `undefined` — `{liveBound: undefined}` would now reset to the default where it used to be a no-op. Checked every caller in `src/`, `tests/`, and the sim harness: `main.ts` passes a real closure and never nulls or omits the key, and all ~60 test callers pass real functions or an explicit `null`. Bare `undefined` never occurs, so the `setLiveBoundCheck` shim is behavior-preserving for every existing caller.

Also confirmed the contract is now genuinely universal across all 15 ports, and that the three claims in the commit message hold: `isLiveBound` really is called without a null guard on the push path, `main.ts` really does only ever set this port, and the wired closure really does degrade to `false` through optional chaining once `crdtLiveViews` is torn down.

Issue #381, opened 2026-08-03
Flaky (~1/8 local): provider-registry teardown-during-hydration leaks an unhandled NoteDestroyedError


Rasbandit:
`tests/crdt/provider-registry.test.ts` → **"genesis edit on A syncs to B and flushes B's disk (no text-verify)"** fails intermittently at roughly **1 run in 8**, locally and reproducibly on `main`.

## Reproduction

```
for i in $(seq 1 12); do bun test tests/crdt/provider-registry.test.ts; done
```

Measured 1/8 on `main` (1baf9ad) and 1/8 on an unrelated feature branch — the branch under test touched only `src/sync.ts` and `src/main.ts`, never `src/crdt/provider-registry.ts`. **Pre-existing, not a regression.**

## Failure

Not an assertion failure — an error escaping the test:

```
NoteDestroyedError: Note was destroyed: n1
  owner: "Note", detail: "n1", noteId: "n1", path: undefined
      at destroy (src/crdt/provider-registry.ts:628:25)
      at destroyAll (src/crdt/provider-registry.ts:646:61)
      at <anonymous> (tests/crdt/provider-registry.test.ts:149:11)
```

Line 628 is where the error is CONSTRUCTED (`e.lifetime.end(new NoteDestroyedError(noteId))`), so that stack is the teardown call chain, not where the rejection went unobserved.

## Diagnosis

`Lifetime.guard` (`src/lifetime.ts:65`) rejects a guarded operation *the instant* the lifetime ends, rather than when the operation settles — deliberately, so a continuation never resumes into dead state. `destroy` ends the lifetime **before** any await, also deliberately, "so in-flight ops are abandoned, not merely checked".

The consequence: at teardown, every in-flight guarded op rejects at once. Any of them that is fire-and-forget (started but never awaited or caught) becomes an unhandled rejection, and Bun attributes an unhandled rejection to whichever test is running — which is why the failure lands on this test rather than on the one that started the op.

Prime suspect is `entry()` (`provider-registry.ts:153`), which awaits `e.lifetime.guard(e.ready)` — the IndexedDB hydration window. A `destroyAll` landing mid-hydration rejects it, and any caller that fires `entry()` without catching leaks it. That window is timing-dependent, which fits the ~1/8 rate.

The two `guard` sites in `applyLocalEdit` (lines 358, 366) are both already awaited inside a `try/catch`, so they are not the leak.

## Fix direction

[...]

PR #374, opened 2026-08-02
fix: repo-review remediation — safety/correctness fixes, dead-code purge, DRY consolidation

Rasbandit:
Remediation of the 2026-08-01 whole-repo review (DRY + code quality + best practices). One PR per the bundling convention; 17 commits, each batch independently green (build + 2,387 tests + biome/eslint/stylelint).

## Behavior fixes (TDD: every one landed with a failing test first)

**Safety**
- **Zombie socket**: `disconnect()` can now cancel an in-flight `openSocketInner` (epoch guard). Previously a backend/vault switch mid-token-fetch resumed into a live socket + heartbeat + reconnect loop the plugin believed dead.
- **Stale beacon token**: `updateConfig`/`setAuthProvider` clear the cached `lastToken`, so a pending telemetry batch can never POST the old backend's JWT to a new origin (dark-launched path, fixed before it ships).
- **`upgrade_url` scheme validation**: the 402 body's URL reaches `window.open` (Electron external-open); non-http(s) schemes from a malicious self-host backend are now dropped at the parse choke point.
- **One sync error boundary**: the four palette commands had no error handling (limit hit = unhandled rejection, nothing on screen). `handleSyncError` now backs all seven entry points; background syncs stay silent on generic failures so offline launches don't toast.

**Correctness**
- **CRLF frontmatter unification**: the editor's `frontmatterPrefixLen` regex diverged from `splitFrontmatter` (the byte-for-byte backend mirror) on CRLF opening fences and trailing-space closing fences — each disagreement is the prefix-offset corruption class. The prefix is now computed FROM the codec, so they can never disagree. The old CRLF test encoded the divergence and was updated to the unified contract (backend is LF-only; verified against `Engram.Notes.Frontmatter.split/1`).
- **Settings progress bar chaining**: `renderContent` bare-assigned the single `onSyncProgress` slot, freezing an open progress modal and leaking a detached-DOM closure after `hide()`. Now capture-chain-restore like `runSyncWithProgress`, with re-render replacing instead of stacking.
- **enrolledIds choke point**: `ProviderRegistry.destroy()/destroyAll()` clear enrollment; the 6 hand-paired `removeDoc`+`reset` sites collapse into `teardownCrdtDoc`.
[...]

PR #382, opened 2026-08-03
refactor(sync): one adoptCreateAck for all three create-ack paths

### LINKED ISSUE #377: refactor(sync): merge the 3 copies of the create-ack flow into adoptCreateAck
Follow-up from the 2026-08-01 repo review (deferred from PR #374).

The "server acked a `crdt_create`" bookkeeping exists THREE times in `src/sync.ts`, and one copy has already leaked a step historically (its own comment admits the queued path "previously LEAKED" the retire step the live path had):

1. `pushFile`'s genesis branch (live path: adopt server id, transfer live mint buffer, seed body, set `CRDT_HEAD_CREATED`, `confirmNoteId`, `flushHeldEditsOnCreateAck`)
2. `applyCrdtCreateAck` (queued path)
3. `recordCrdtGenesisPushed` (batch path, no live-transfer)

Extract one `adoptCreateAck(localId, serverId, path, opts)` used by all three. The flows differ subtly (live-transfer vs not, seed source, echo-window behavior), which is both why a fourth divergence is a matter of time and why the merge needs careful supervised work with e2e coverage — deliberately not done in the overnight remediation.

Related detail worth preserving in the merge: PR #374 made the genesis branch open the echo-cooldown window when the seed shipped (`success = true`); the post-create-throw exit conservatively leaves it closed (see comment at the site).

Rasbandit:
Closes #377.

> **Stacked on #380** (the #376 prereqs) — this branch's base is `refactor/sync-state-mutators-crdt-ports`, because #377 touches exactly the code prereq 1 rewrote. Review/merge #380 first; retarget to `main` once it lands.

## The drift

The "server acked our `crdt_create`" bookkeeping existed THREE times, and one copy had already leaked a step historically — its own comment admits the queued path missed the mint-retire the live path did.

| Step | live (`pushFile`) | queued (`applyCrdtCreateAck`) | batch (`recordCrdtGenesisPushed`) |
|---|---|---|---|
| adopt id → `noteIdMap` | ✅ | ✅ | ✅ |
| `setCrdtHead` sentinel | ✅ | ✅ | ✅ |
| `confirmNoteId` | ✅ | ✅ | ✅ |
| `flushHeldEditsOnCreateAck` | ✅ | ✅ | ❌ omitted |
| echo baseline stamp | **after** the flush | **before** the flush | unconditional |

Two copies of a five-step sequence that already lost a step once, plus two different orderings of the last one. A fourth divergence was a matter of time.

## The merge

All three now call one `adoptCreateAck(effectiveId, path, consumed, opts?)`.

**The ADOPT half deliberately stays with each caller** — it legitimately differs by path: a live editor buffer to transfer and a mint doc to retire, a queued disk seed, or nothing at all for a batch note that never minted a local doc. Only the shared tail merges, which is where the drift actually was.

### Ordering, now stated once at the helper

- **`confirmNoteId` before the flush.** The confirm opens Task 1's `canSendLive` gate, so the flush is what finally ships the held edits. Reversing them silently strands the edits.
- **The baseline before the awaited flush.** A **deliberate tightening of the live path**, which stamped it after: `flushHeldEditsOnCreateAck` is awaited, so stamping after leaves a window where the create's own broadcast returns before the baseline that suppresses it exists. The queued path already had the safe order; the live path now matches it.

### The one honest difference

The batch path passes `flushHeld: false` — it seeds no local doc (content ships inline in the batch frame), so it has no gated updates to flush. Now a **named argument instead of a silently missing line**. `recordCrdtGenesisPushed` becomes async for it; its single caller already sits in an async loop.

## Preserved

The post-create-throw exit still passes a null baseline, leaving the echo-cooldown window closed conservatively (an unsuppressed self-echo is absorbed by the hash-skip dedupe) — the behavior #374 called out at that site. The queued path keeps its own `noteIdMap` remap up front rather than leaving it to `adoptCreateAck`: the ViewPlugin re-resolves path → id on its next update, which can land during the awaited transfer/retire.

## Tests

New `tests/crdt-adopt-create-ack.test.ts` pins the contract — the ordered core, baseline-before-flush, confirm-before-flush, a null `consumed` stamping no baseline while still running the rest, `flushHeld:false` skipping the flush, merge-at-stamp-time, and both writes keyed by the same normalized path.

## Verification

- `bun run build` (tsc + esbuild) clean
- `bun test` — 2409 pass, 1 skip, 0 fail (includes the randomized `sim/random` convergence tier)
- `biome ci src tests`, `lint:obsidian`, `lint:css` clean

The issue asks for e2e coverage; the genesis/create-ack flows are exercised by the Obsidian e2e suite, which runs in CI on this PR.

**Unrelated flake found while verifying:** `tests/crdt/provider-registry.test.ts` fails ~1/8 on `main` too, in a file this branch never touches. Diagnosed and filed as #381 rather than folded in here.


Rasbandit:
### Code review

Found 1 issue, fixed in 22b756d:

1. The consolidated `adoptCreateAck` doc misattributes which step opens Task 1's live-send gate. It says `confirmNoteId` opens `canSendLive` — but `main.ts` wires `canSendLive: (id) => this.syncEngine.hasServerNote(id)`, and `isNoteConfirmed`'s own doc in the same file explains why: `confirmedNoteIds` is session-scoped and cleared on every WS reconnect, so the gate was deliberately moved off it onto the `crdtHead` oracle. The step that actually opens the gate is `setCrdtHead`'s sentinel flip.

https://github.com/engram-app/Engram-obsidian/blob/35c138880685f713f7d6d612d8b8cfb230933181/src/sync.ts#L899-L902

Contradicted by the wiring:

https://github.com/engram-app/Engram-obsidian/blob/35c138880685f713f7d6d612d8b8cfb230933181/src/main.ts#L2105-L2107

and by the field's own doc:

https://github.com/engram-app/Engram-obsidian/blob/35c138880685f713f7d6d612d8b8cfb230933181/src/sync.ts#L829-L835

The code is correct — `setCrdtHead` runs first and both run before the flush — but the comment pointed a future reader at the wrong load-bearing line. Someone preserving "confirm before flush" while moving `setCrdtHead` after it would ship into a still-closed gate. In this file the comments are the invariant record, so a misdirecting one is worth fixing. The wording was inherited from the three pre-refactor call sites; consolidating them into one authority is what made it worth getting right. The test pinned the weaker ordering for the same wrong reason, and now traces the oracle flip instead.

Checked and found clean: the `consumed` threading in `applyCrdtCreateAck` (the two seed routes are genuinely mutually exclusive via the `!transferredLiveContent` file gate), `recordCrdtGenesisPushed` becoming async against its single already-awaited call site, the `flushHeld: false` carve-out against the original which never flushed on that path, and the #374/#377 concerns about preserving the echo-cooldown window on the post-create-throw exit.

Not flagged: the `sync.ts:NNNN` line-number self-references drift further out of date here, but 4 of the 5 already pointed at blank or unrelated lines on `main`, so that is pre-existing.

Rasbandit:
### Code review

Follow-up pass over `22b756d` and the merge `abb13e7` (both unreviewed at the time of the previous review).

No issues found.

Every factual claim in the rewritten doc comment was checked against the code: `canSendLive` is wired at `src/main.ts:2102` to `hasServerNote` and nowhere else in the codebase; `clearConfirmedNoteIds` is called from `channel.onStatusChange`'s connected branch, so `confirmedNoteIds` genuinely dies on every reconnect; `hasServerNote` reads exactly the field `setCrdtHead` writes; and `adoptCreateAck` orders oracle flip before confirm before the awaited flush, matching both the doc and the test.

The test change is strictly additive — the trace harness gained a hook and each sequence assertion gained the new element in its actual execution position. Nothing was removed or loosened, and the original confirm-before-flush assertion is retained as a secondary pin.

Merge `abb13e7` verified clean: each parent's diff-to-merge equals the other parent's unique diff, so no hunk was lost; the `liveBound` fix and the `adoptCreateAck` refactor coexist intact; the committed `main.js` rebuilds byte-identical (not stale).

One nuance, not worth a change: the comment calls `confirmNoteId` a feed for "in-session bookkeeping", and its two readers (`refireEnrollmentOnFirstConfirm`, `healNoteOnOpen`) do drive real control flow. The claim being made — that it is not the live-send gate — is correct.

PR #384, opened 2026-08-04
docs(context): what opens the live-send gate

Rasbandit:
Extends `docs/context/crdt-pull-gated-by-create-ack.md` rather than adding a new doc.

That doc already documents the gate wiring correctly (`canSendLive: (id) => syncEngine.hasServerNote(id)`). A second doc restating it would be a fourth copy of the same fact — precisely the drift class #377 existed to end.

## What's added

The **write** side of the same gate. The existing content covers the pull side (a handshake wrongly held); this covers what opens the gate and the ordering the create-ack paths must keep.

- **`setCrdtHead` opens `canSendLive`, not `confirmNoteId`.** `confirmedNoteIds` is session-scoped — cleared from `channel.onStatusChange`'s connected branch on every reconnect — which is exactly why the gate was deliberately moved onto the `crdtHead` oracle. Its two remaining readers (`refireEnrollmentOnFirstConfirm`, `healNoteOnOpen`) are not the live-send gate.
- **The trap.** Inline comments at all three create-ack sites said the gate was opened by "confirm it, then flush". The code was always correct (`setCrdtHead` ran first), but the prose pointed at the wrong load-bearing line — someone preserving "confirm before flush" while moving `setCrdtHead` after it would ship into a closed gate. Caught reviewing #382, fixed in `22b756d`.
- **The three create-ack callers**, why the ADOPT half deliberately stays per-caller, and the `flushHeld: false` batch carve-out.
- **The two load-bearing orderings** `adoptCreateAck` keeps, including why the echo baseline now stamps before the awaited flush — a behavior change, not just a refactor.

Also flags that this doc's `sync.ts:NNNN` references predate #380/#382, and that most were already stale before that: grep the symbol, not the line.

Doc-only change; `CLAUDE.md` gains one pointer line for the new trigger.

Refs #377

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -175,5 +175,6 @@ Required status checks on `main`: `build-and-test`, `version-check / version-che
 - Prod `auth-failure-burst` alert traced to our own client (`Bearer ` with an empty token logs `reason=no_auth`, not `signature_error`; unlinked installs loop 401s and the log push re-reports them) → `docs/context/empty-bearer-no-auth-401-loop.md`
 - Convergence sim tier fidelity gaps (`tests/sim/` — differential gate pays rent; seeded random-op suite does NOT converge, kept as a runnable tool not a test) → `docs/context/crdt-convergence-sim-fidelity-gaps.md`
 - A note that logs `re-handshake fired` and then goes silent forever (the create-ack gate swallowing syncStep1; plus how to read `ci-debug` client logs without misordering them) → `docs/context/crdt-pull-gated-by-create-ack.md`
+- Touching create-ack bookkeeping, or wondering what actually opens the live-send gate (`setCrdtHead`/`hasServerNote`, NOT `confirmNoteId` — and the ordering `adoptCreateAck` must keep) → `docs/context/crdt-pull-gated-by-create-ack.md`
 
 @/home/open-claw/documents/code-projects/ops-agent/docs/self-updating-docs.md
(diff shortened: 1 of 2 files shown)


Rasbandit:
### Code review

One issue, fixed in 81a2984.

1. The new section said `setCrdtHead` opens the gate, which is incomplete. `hasServerNote` resolves the id through the noteIdMap **first** and returns false on a miss, so the gate needs BOTH a map entry and a head on that path:

https://github.com/engram-app/Engram-obsidian/blob/9d1b5c7d94b6d0e5d6e1e3fd2f6a1c9e0b7d8a3f/src/sync.ts#L2053-L2058

An id absent from the map is gated exactly as hard as one with no head. That is also why `adoptCreateAck` does `noteIdMap.set` as its first statement, before the oracle flip — flipping the head for a path whose id is not yet mapped leaves the gate shut anyway. Since precision about this gate is the entire point of the doc, an incomplete condition is worth correcting rather than leaving.

Every other factual claim was checked against the code and holds:

| claim | verified |
|---|---|
| `canSendLive: (id) => syncEngine.hasServerNote(id)` | `src/main.ts:2102`, and it is the only wiring of `canSendLive` in the repo |
| `confirmedNoteIds` cleared on every reconnect | `clearConfirmedNoteIds()` called from `src/main.ts:1912`, inside `onStatusChange`'s connected branch |
| `hasServerNote` reads what `setCrdtHead` writes | `getCrdtHead(path) != null`, and `setCrdtHead` writes `crdtHead` via `patchSyncedRow` |
| `adoptCreateAck` ordering | `noteIdMap.set` → `setCrdtHead` → `confirmNoteId` → baseline → awaited flush, matching the doc |
| `confirmedNoteIds`' two remaining readers | `refireEnrollmentOnFirstConfirm`, `healNoteOnOpen` — neither gates live sends |

PR #385, opened 2026-08-04
fix(deps): clear three new audit advisories

Rasbandit:
`bun audit` — the "Security audit" step of the **required** `build-and-test` check — went red on `main`, which blocks every open PR (#379, #384). This unblocks them.

All three are real advisories with real fixed versions, not merged-range artifacts. Checked individually, per the explicit warning in `docs/context/bun-audit-brace-expansion-false-positive.md` ("do not assume every `bun audit` high is a false positive").

| advisory | sev | vulnerable | our state | fix |
|---|---|---|---|---|
| [GHSA-rgw5-rvv9-x895](https://github.com/advisories/GHSA-rgw5-rvv9-x895) brace-expansion | high | `>=4.0.0 <5.0.9` | override `^5.0.8` — **inside the range** | `^5.0.9` |
| [GHSA-7p8r-x3mc-p8w7](https://github.com/advisories/GHSA-7p8r-x3mc-p8w7) fast-uri | high | `>=3.0.0 <3.1.5` | unpinned | `^3.1.5` |
| [GHSA-fxqj-rqcc-2cmp](https://github.com/advisories/GHSA-fxqj-rqcc-2cmp) postcss | moderate | `<=8.5.22` | unpinned | `^8.5.23` |

**The brace-expansion one is worth noting:** this is a *new* advisory (DoS via unbounded intermediate arrays, bypassing the CVE-2026-14257 mitigation), not the GHSA-mh99 ReDoS that our `^5.0.8` override was originally added for in #324. That override was correct for the old advisory and lands squarely inside the new one's range.

Reach:
- fast-uri — `eslint`/`eslint-plugin-obsidianmd`/`stylelint` → `ajv` → `fast-uri`
- brace-expansion — `eslint`/`typescript-eslint`/`eslint-plugin-obsidianmd` → `minimatch` → `brace-expansion`
- postcss — `stylelint` → `postcss-safe-parser` → `postcss`

All dev-toolchain only, none imported from `src/`, so no runtime or bundle impact.

The doc's paired-override trap (bumping `brace-expansion` alone breaks eslint's `minimatch@3`, which expects the bare-function export) does **not** bite here — `minimatch` is already pinned at `^10.0.1`, and both eslint and stylelint still pass, so no intermediate consumer needed bumping alongside these.

## Verification

- `bun audit` — No vulnerabilities found
- `lint:obsidian` (eslint), `lint:css` (stylelint), `biome ci`, `lint:lockfile` all pass
- `bun run build` clean
- `bun test` — 2412 pass, 1 skip, 0 fail

Lockfile portability check passes, so `bun.lock` stays contributor-safe.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -41,10 +41,12 @@
     },
   },
   "overrides": {
-    "brace-expansion": "^5.0.8",
+    "brace-expansion": "^5.0.9",
+    "fast-uri": "^3.1.5",
     "js-yaml": "^4.3.0",
     "minimatch": "^10.0.1",
     "picomatch": "^2.3.2",
+    "postcss": "^8.5.23",
   },
   "packages": {
     "@babel/code-frame": ["@babel/code-frame@7.29.7", "", { "dependencies": { "@babel/helper-validator-identifier": "^7.29.7", "js-tokens": "^4.0.0", "picocolors": "^1.1.1" } }, "sha512-Aup7aUOfpbAUg2ROOJN6Iw5f9DMBlzu0mIkm/malLQFN/YQgO48wCj0Kxa3sEHJvPVFg7siR+qRInwXd2qhQKw=="],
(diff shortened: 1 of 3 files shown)


Rasbandit:
### Code review

Post-merge review — this PR was merged without one, because it was blocking every other PR. That was the wrong call regardless of urgency, so here is the pass, done after the fact.

**No issues found.** The three override bumps are correct and the advisories are real.

One claim in the commit message needed checking: it asserts "no runtime or bundle impact", yet the commit changes `main.js` by 104 insertions / 104 deletions. Verified — **every one of the 208 changed lines is a `node_modules` path comment**, and there is no functional change:

```
../../node_modules/diff-match-patch/index.js  ->  node_modules/diff-match-patch/index.js
../../node_modules/yaml/browser/dist/...      ->  node_modules/yaml/browser/dist/...
```

So the claim holds. But the reason it changed is worth recording, because it is a hazard independent of this PR:

**The committed `main.js` embeds the builder's `node_modules` resolution path.** A worktree that resolves `node_modules` upward to the canonical checkout bakes `../../node_modules/...` into the bundle; one with a real local `node_modules` bakes `node_modules/...`. `main`'s artifact carried the `../../` form before this commit, meaning it was last built from a worktree without local deps. Running `bun install` here normalized it.

Consequences worth being aware of:
- Bundle diffs churn by ~200 lines depending on *where* the build ran, which is noise that can mask a real change in review.
- Whether `main.js` is "current" is not a stable question — it depends on the build environment, not just the sources. That undercuts the "rebuild and check `git status`" staleness check used elsewhere in this stack.

Filed separately rather than fixed here. Not a correctness problem — the emitted code is identical either way.

Also confirmed: `minimatch` is already pinned at `^10.0.1`, so the paired-override trap from `docs/context/bun-audit-brace-expansion-false-positive.md` does not apply; eslint and stylelint both pass. Lockfile portability check passes, so `bun.lock` stays contributor-safe.

PR #404, opened 2026-08-07
refactor(sync): one op-log pager for replay + enumerate

### LINKED ISSUE #378: refactor(sync): one op-log pager (walkOpLog) for enumerate + replay
Follow-up from the 2026-08-01 repo review (deferred from PR #374).

`enumerateServerState` re-implements `runSeqReplayOnce`'s op-log pagination in `src/sync.ts`: two fetch/cursor-advance loops, two stuck-cursor guards, and two copies of the page size (500) and loop ceiling (100_000) that must be kept in sync by hand.

Fix: extract one `walkOpLog(onRow)` pager both call (or route `enumerateServerState` through `catchupViaSeqReplay({enumerateOnly: true})`, which already exists for a read-only walk). Name the two magic numbers once.

Touchy replay code: keep the exclusive-replay semantics (`catchupViaSeqReplayExclusive`) and the coalesced-replay empty-set caveat documented at the replace-remote call site intact.

Rasbandit:
Closes #378

## What

`enumerateServerState` re-implemented `runSeqReplayOnce`'s op-log pagination: two fetch/cursor-advance loops, two stuck-cursor guards, and two copies of the page size (500) and loop ceiling (100_000) that had to be kept in sync by hand.

Extracted `walkOpLog({seq, id, onRow, onPage})`. It owns the mechanics both walkers duplicated; each caller keeps only its own collection and side effects. The two magic numbers are named once as `OP_LOG_PAGE_SIZE` / `OP_LOG_MAX_PAGES`.

## Why unifying the cursor advance is safe

The two walkers advanced differently — enumerate used the backend's `next_seq`/`next_id`, the replay used the max `{seq, id}` of the rows it saw. These are provably the same value, not assumed to be: the backend derives `next` from the **last row** of a `{seq, id}`-sorted page (`merged_changes_page`, engram `lib/engram/sync.ex:45-58`). The pager keeps the replay's row-derived version verbatim, so the touchy path is unchanged.

## What deliberately still differs

Fetch-error policy. Errors propagate out of the pager and each caller decides:

- the **replay** swallows them, keeping the pages it already applied and their persisted cursor
- the **preview** lets them surface, rather than rendering a plan off a short walk

## Untouched

Exclusive-replay semantics (`catchupViaSeqReplayExclusive`) and the coalesced-replay empty-set caveat documented at the replace-remote call site.

## Tests

Four new tests pin the contract the two callers must **share** (page size, stuck-cursor guard on both paths) and the one they must **not** (fetch-error policy).

The stuck-cursor tests are self-limiting: the mock throws once the walker over-fetches, so a removed guard fails in milliseconds instead of spinning the 100k-page ceiling.

Verified non-vacuous by mutation, before and after:
- **before** — mutating each of the three duplicated guards independently killed its own test
- **after** — a *single* mutation to the unified pager fails **both** callers' tests, which is the property the refactor buys

Full suite green: 2439 pass / 0 fail. `tsc`, biome, eslint, stylelint, lockfile check all clean.

Net +14 non-comment lines — this trades a little plumbing for removing the hand-sync hazard; it is not a raw-line deletion.

https://claude.ai/code/session_016D4V5B8Yixkk6TjG8Yc4B2

### DIFF
--- a/main.js
+++ b/main.js
@@ -18969,7 +18969,13 @@ var BINARY_EXTENSIONS = /* @__PURE__ */ new Set([
   webm: "video/webm",
   zip: "application/zip",
   canvas: "application/json"
-}, PUSH_BATCH_SIZE = 10, _SyncEngine = class _SyncEngine {
+}, PUSH_BATCH_SIZE = 10, OpLogFetchError = class extends Error {
+  constructor(cursorSeq, cause) {
+    super(`op-log fetch failed at cursor=${cursorSeq} \u2014 ${errMsg(cause)}`);
+    this.cursorSeq = cursorSeq;
+    this.name = "OpLogFetchError";
+  }
+}, _SyncEngine = class _SyncEngine {
   constructor(app, api, settings, saveData, time = new DefaultTimeProvider()) {
     this.app = app;
     this.api = api;
@@ -20997,7 +21003,12 @@ var BINARY_EXTENSIONS = /* @__PURE__ */ new Set([
     if (!fetchPage) return;
     let cursor = opts.seq, cursorId = opts.id;
     for (let page = 0; page < _SyncEngine.OP_LOG_MAX_PAGES; page++) {
-      let pageStartSeq = cursor, pageStartId = cursorId, resp = await fetchPage(cursor, _SyncEngine.OP_LOG_PAGE_SIZE, cursorId);
+      let pageStartSeq = cursor, pageStartId = cursorId, resp;
+      try {
+        resp = await fetchPage(cursor, _SyncEngine.OP_LOG_PAGE_SIZE, cursorId);
+      } catch (e) {
+        throw new OpLogFetchError(cursor, e);
+      }
       for (let c of resp.changes)
         await opts.onRow(c), this.cursorAdvances(
           typeof c.seq == "number" ? c.seq : null,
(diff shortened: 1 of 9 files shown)


Rasbandit:
## Self-review pass — two defects found and fixed (f60126c)

Reviewed this diff at high effort before asking for a merge. It turned up one bug I introduced and one that was already there and worse.

### 1. The refactor's `try` was wider than the loop it replaced *(introduced here)*

The original only wrapped the `crdtCatchupSince` fetch. My version wrapped the whole walk, which meant a `saveData` failure in the per-page cursor persist got swallowed and logged as `seq-replay: fetch failed at cursor=N` — wrong cause, and previously it propagated.

`walkOpLog` now tags transport failures as `OpLogFetchError`, and the replay swallows **only** those. Errors from its own `onRow`/`onPage` work stay loud.

### 2. A delete decision could run on a partial server set *(pre-existing)*

This is the one that matters. A walk cut short mid-feed still returned `ran: true` with **partial** `serverIds`/`serverAttachmentPaths`. `catchupViaSeqReplayExclusive` only guards the *coalesced* case (empty sets) — partial-but-non-empty sails straight past it into `pullAll({ deleteLocalExtras: true })`, where every note the walk never reached looks server-absent and gets trashed.

Concretely: 500-note vault, socket drops after page 1, everything past page 1 goes in the trash. The unwired-socket path (`!crdtCatchupSince` → `ran: true` + empty sets) was the same hazard with an empty set.

`runSeqReplayOnce` now reports `{applied, complete}`; `catchupViaSeqReplay` propagates `complete` (one short pass taints the accumulated sets — never ORed back to true); the exclusive replay returns `null` when incomplete, so destructive callers take the abort path they **already implement**.

### Tests

Both new tests were watched failing first:
- a persist failure must not be swallowed
- an exclusive replay must refuse a walk that stopped early

### On the test-double churn

Stubs asserting `ran: true` now also say `complete: true` — that is what they always meant. Two fixtures changed *premise*, not assertion:

- the unwired-fetcher test now expects `complete: false`: empty sets from a walk that never ran describe our ignorance, not the server
- **`pullAll` delete-local wired no server at all**, so it was passing *through* the degenerate empty-set path this function's own docstring calls the data-loss bug. It now wires a socket that reports an empty server over a complete walk, so the engine knows rather than assumes.

Worth noting the polarity: a stub that forgets `complete` reads `undefined` and fails **closed** — it refuses to trash.

2441 pass / 0 fail. Build, biome, eslint, stylelint clean (verified on exit codes, not piped output).

PR #449, opened 2026-08-18
fix(sync): probe the empty-trust guard by id, not by event.path

Rasbandit:
Fixes the data-loss path in engram-app/engram#1377. Filed on the backend, but the fix belongs here.

## The bug

`content_hash` is a per-user HMAC, so `hmac("")` is ONE constant for the whole vault. `applyChange` learns it from any op-log row carrying `content: ""`, and the ingress guard in `handleStreamEvent` then trusts a broadcast's inline `""` whenever the hash matches.

That makes `hash -> content` non-injective in exactly the way the guard assumes it is not. Matching `H_empty` proves "this hash means empty". It does not prove "this note is empty".

engram#1375 turned that latent weakness into observable loss:

1. A note whose last checkpoint was empty but which now has uncheckpointed ops keeps `H_empty` as its STORED hash. #1375 deliberately serves the facade's hash rather than recomputing.
2. The change feed therefore delivers that note's REAL body under `H_empty`.
3. A meta-projected cascade broadcast (the folder-rename shape the guard was written for) then arrives carrying `content: ""` with that same `H_empty`.
4. It matches the learned value, so the guard trusts it and materializes a 0-byte file over the body the feed just delivered.

Not a regression from #1375. Before it, the local copy was already empty, so the same broadcast overwrote empty with empty. #1375 makes the body arrive, which is what turns a latent weakness into real loss.

## Why not note-scope the learn

The issue's first suggestion was to make the empty-hash trust note-scoped. That does not fix it, and it deletes the optimization: trusting `""` only for a note already known empty buys nothing, because writing `""` over an already-empty note is a no-op. It would also break the existing `tests/sync-empty-content-distrust.test.ts` case that deliberately teaches from `a.md` and trusts a broadcast for `b.md`.

The issue's other suggestions were considered and rejected:

* Gating on `crdt_head` requires the manifest, which #1401 is actively retiring. Building a new dependency on an oracle we are removing is the wrong direction.
* Having the backend omit `content_hash` on meta projections makes things worse, not better. The guard reads `content === "" && content_hash && ...`, so dropping the hash means the guard never fires and the 0-byte body materializes anyway. It is also a cross-repo pairing deadlock.

## The fix

The discriminator is not WHICH note. It is whether the write DESTROYS bytes. Trusting inline `""` is only ever harmful against a local file that has some.

```ts
event.content_hash !== this.emptyContentHash ||
  (this.app.vault.getFileByPath(normalizePath(event.path))?.stat?.size ?? 0) > 0
```

`stat.size` answers that without reading the file, and distrust routes to the same op-log seq-replay that already heals this path. A note that genuinely went empty still converges via its authoritative row, at the cost of one replay.

The vault-wide learn is kept. For a note with no local bytes the trust is free and the round-trip saving is real.

## Testing

New case in `tests/sync-empty-content-distrust.test.ts`: teach `H_empty` from a genuinely empty note, give the victim a local file with `stat.size > 0`, then deliver the poisoned broadcast. Asserts no 0-byte write lands on the victim by either `create` or `modify`, and that the catch-up replay ran instead.

Verified honest: stashing only the `src/sync.ts` change turns the new test red, restoring it turns it green.

Gates, all green locally:

* `bun test` 2779 pass, 0 fail (2778 before)
* `biome ci`, `tsc` build, `lint:obsidian`, `lint:css`

Refs engram-app/engram#1377

https://claude.ai/code/session_01VmRPf5NEbZi1zqnjwKgBT5


Rasbandit:
## Corrected after review. The original claim was wrong.

Two findings invalidated the first version, both confirmed against the code:

**The guard was inert on the CRDT path, i.e. on the notes #1377 is actually about.** Both inline-apply sites in the CRDT branch are already gated on `!getAbstractFileByPath(np)` (`sync.ts:6089`, `:6105`). A file with bytes makes that non-null, so `applyOp` was already skipped and control had already fallen through to `materializeRelocated` / replay. A condition keyed on "a file with bytes exists" cannot change that branch by construction.

**The test only went red because its vault mock was impossible.** It stubbed `getFileByPath` to return a file while `beforeEach` left `getAbstractFileByPath` at null. Real Obsidian cannot answer "a file here" to one and "nothing here" to the other, since the latter is the former widened. The CRDT branch keys its gate on the accessor left null, so that inconsistency alone flipped behaviour. My stash-and-watch-it-go-red check was measuring the mock, not the bug.

## What the PR is now

Narrower, and proven. Probe the ID's current local path instead of `event.path`:

```ts
const emptyTrustPath =
    (event.id ? this.noteIdMap?.pathForId(event.id) : null) ?? event.path;
```

A folder-rename cascade, the shape that produces these meta-projected broadcasts, names the NEW path while the bytes are still at the OLD one. Probing `event.path` finds nothing, scores zero bytes, trusts the inline `""` and creates a 0-byte file at the new path. Probing by id sees the bytes and refuses.

## Test changes

* `registerFile` keeps both vault accessors consistent and returns a **real `TFile`**. Every write in `sync.ts` is gated on `instanceof TFile`, so a plain object routes around all of them and makes a "no empty write landed" assertion pass vacuously. That was a second, independent way the original test lied.
* **Deleted the non-rename case.** With a consistent mock and a real TFile it passes identically with and without the diff, so it implied coverage it did not provide. Something else already protects that path.
* The remaining case fails without the fix and passes with it. Verified by stashing only `src/sync.ts`: 5 pass / 1 fail without, 5 pass / 0 fail with.

## This does not close #1377

The reported scenario runs through the CRDT branch, which the file-absence gate already covers. #1377 needs re-verification rather than closing on this evidence, and I have said so on the issue.

Also still open from the review and not addressed here: a stale-`H_empty` broadcast on a device that does not yet have the note still materializes a 0-byte file at first delivery, and the CAS seed then stamps it as converged. That is the same "0-byte file sticks" class at a different moment, and it wants its own change.

PR #450, opened 2026-08-18
fix(crdt): stop reporting a refused index frame as delivered

### LINKED ISSUE #433: index-crdt: a rejected index frame has no recovery path
Follow-up from the adversarial review of #431 (finding F5, severity 75).

`NoteChannel.sendIndexCrdt` is fire-and-forget: it registers no pending-reply entry and returns `true` unconditionally once `this.send(...)` has run. The server's error replies — `rate_limited` and `index_frame_rejected` (`engram/lib/engram_web/channels/crdt_channel.ex:313-324`) — land in the generic `status === "error"` branch, miss the `note_not_found` special case, and surface as a misleading *"Channel join error on crdt:..."*.

**Why this is worse for the index room than for notes.** A rejected NOTE frame is survivable: the doc still holds the state and the next reconnect's syncStep1 re-diffs it. The index room's ops are already in the local doc and already committed out of the overlay, so a later syncStep1 will not re-offer them as a fresh update. **Every dropped index frame is a permanently lost path→id claim.**

Live triggers:
- **Rate limit.** `crdt_channel.ex:1295` classifies any non-step1 frame as `:edit`, sharing the 240-per-10s per-device budget with note content edits. `adoptCreateAck` (`sync.ts`) calls `noteIdMap.set` on every create-ack, so a bulk import emits an index frame per note *on top of* its content frames. ~300 files in a 10s window starts getting denied.
- **Size.** Frames above the 5 MB decoded ceiling are rejected outright — see the sibling issue on the seed frame.

Also related, same class: `NoteChannel.send` silently no-ops when `readyState` is `CLOSING`/`CLOSED`, while `sendIndexCrdt` still returns `true`. `crdtJoined` is only reset in `onclose`, so there is a real half-open window (review finding F8, severity 50).

## Options
1. Await the reply for index frames and re-stage on error.
2. Re-buffer into the provider on an error reply.
3. Keep a small outbound journal of unacked index updates, replayed on rejoin.

Whichever wins, "reported delivered but never applied" has to stop being a silent outcome.

Refs #362

Rasbandit:
Closes #433.

## What was broken

Two paths reported an index frame delivered when the server never applied it.

**Half-open socket (F8).** `NoteChannel.send` silently no-ops unless `readyState` is `OPEN`, and `crdtJoined` is only reset in `onclose`. `sendIndexCrdt` returned `true` regardless. Since `NoteProvider.broadcast` buffers only on a `false` return, nothing held the frame.

**Refused frames (F5).** The server replies to every `crdt_index_msg`, either `{:ok, %{}}` or an error. Those replies went nowhere. `rate_limited` and `index_frame_rejected` fell through to the generic error branch and surfaced as `Channel join error on crdt:...`, so a dropped identity claim read as a connection fault and nothing re-offered the frame.

## Correcting the issue's severity

The issue rates this permanent loss, on this premise:

> The index room's ops are already in the local doc and already committed out of the overlay, so a later syncStep1 will not re-offer them as a fresh update. Every dropped index frame is a permanently lost path->id claim.

That premise does not hold, and it is worth correcting because it drove the severity-75 rating.

`y_ex` answers an inbound syncStep1 with `encode_sync_step1_response_v1`, which is a **[syncStep2, syncStep1] pair**, not a bare diff. `note-provider.ts:172` already documents exactly this: *"The server answers every inbound syncStep1 with a fresh [syncStep2, syncStep1] pair."* So on rejoin the client reads the server's syncStep1, diffs it against the local doc, and sends back a syncStep2 carrying everything the server lacks. The refused claim never left the local doc, so it is re-offered.

**A rejoin heals it.** What is real is the window before that rejoin: the server answers for a path it does not own, and `getOrMint` can mint a duplicate id in the meantime. That is worth fixing, but it is a bounded window, not permanent loss.

## The fix

* `send` returns whether the frame reached the transport. `sendIndexCrdt` honours it, so a half-open socket routes into the existing offline buffer with its kind preserved. No new recovery machinery.
* In-flight index frames are tracked by ref. The server replies to every one, so entries drain on reply; they are cleared on close, where the rejoin handshake takes over.
* An error reply calls `onIndexFrameRejected(b64, reason)`, wired in `main.ts` to `IndexRoom.requeue`, which puts the frame back in the provider's offline buffer.

`requeue` deliberately does **not** retry immediately. A refusal is usually a rate limit, and re-sending into one is precisely how the 2026-07 re-handshake storm worked. It rides the next connect flush instead.

Kind is `"op"` for requeued index frames: `IndexRoom`'s own wiring notes that the index room has no create-ack gate, so handshake and op frames take the same path out and the classification cannot strand anything the way it could on a note room.

## One test change, called out deliberately

`tests/source-compliance.test.ts` exempts one exact wire-send shape and is intentionally brittle, so that "any wrapper, extra argument or reordering stops matching and is caught". Reading the send's boolean result breaks that match.

The exemption is widened by exactly one thing: an optional `const <name> =` prefix. The exact wire shape is still pinned, so a wrapper around the payload, an extra argument or a reordering is still caught. The note-frame exemption (`wireStatement`) is left untouched, since no note-frame caller needs the boolean and widening it too would exempt more for free.

## Testing

Four new cases in `tests/channel-crdt.test.ts`:

* `sendIndexCrdt` returns false when the socket is no longer OPEN, and writes nothing
* an `index_frame_rejected` reply hands the exact frame back to be re-offered
* a `rate_limited` reply does the same
* an `ok` reply drops the in-flight entry, so a later reply on that ref finds nothing

Gates, all green locally:

* `bun test` 2787 ran, 0 fail (baseline on this branch point is 2783)
* `biome ci`, `tsc` build, `lint:obsidian`, `lint:css`

Closes #433

https://claude.ai/code/session_01VmRPf5NEbZi1zqnjwKgBT5


Rasbandit:
## test_27 investigation

`test_27_empty_note_sync` failed once on this PR (content written to an empty note never reached the server within 120s). Investigated rather than rerun-and-shrug, because this PR changes how note-content delivery reports success.

**Did not reproduce.** 4/4 passes locally against this exact branch build, against the CRDT stack, ~24s each. The CI rerun then passed too.

**Not claiming that clears it.** A rerun-pass proves non-determinism, not innocence. So here is the mechanism I could not rule out, written down so a reviewer can weigh it:

`sendCrdt` now returns `false` on a half-open socket. `wiring.ts` keys `unsentDocIds` on that boolean, and that set drives `reEnrollUnsent()` — one STEP1 per tracked doc on rejoin. More docs marked unsent therefore means more handshake traffic on reconnect, and `crdt-wrong-mint-cross-file-overwrite.md` names handshake rate-limit starvation as a real trigger in this codebase. It would also be load-dependent, which fits a failure that will not reproduce in isolation.

What lowers the concern:

* `addUnsent` is bounded at `MAX_UNSENT_DOCS` (500) with oldest-eviction, so the burst cannot grow without limit.
* The new refusal cause is the same class the set already exists for ("the frame did not reach the server"). This widens an existing door rather than opening a new one.
* Before this change the frame was silently dropped in that window with nothing to re-offer it, which is strictly worse than one extra STEP1.

Also worth noting the suite was broadly unstable in the same window: `feat/genesis-seed-detached` was failing different tests, and #1420 (docs-only) went red on `test_deaf_note_survives_handshake_rate_limit`.

**Added a guard rather than a fix:** `sendCrdt returns false when the socket is no longer OPEN`, mirroring the index-frame case. It pins the behaviour this PR changes, so a future regression fails loudly.

If test_27 recurs on this branch, the STEP1-burst theory above is where I would start.

PR #467, opened 2026-08-24
fix(sync): close pushFile's TOCTOU reentrancy race

Rasbandit:
## Summary
- `main` on the backend repo is currently broken: `test_97_first_sync_single_lineage`, `test_98_resync_to_fresh_vault_single_lineage`, and `test_100_frontmatter_echo_single_lineage` fail deterministically on every CI run since Engram#1455 merged (2026-08-24), each showing notes with multiple independent Yjs lineages ("worst: 7 clients").
- Traced it to `pushFile`'s reentrancy guard: `this.pushing.has(path)` is checked once at entry, but two `await`s (the echo-hash read, `acquirePushSlot`) sit between that check and `this.pushing.add(path)`. Two overlapping `fullSync()` sweeps for the same never-before-synced file — nothing serializes `fullSync()` calls — can both pass the stale check before either claims the slot, and both push. `test_77`/`97`/`98`/`100`'s own retry loops produce exactly this shape (`while (...) { await trigger_full_sync(); ... }`, never awaiting the previous call's completion).
- For a brand-new CRDT note, two concurrent pushes means two independent throwaway-Y.Doc genesis lineages can land on the same server row — the doubling these tests catch.
- Fix: re-check the guard immediately before the claim, with nothing async in between (single-threaded JS makes that pairing atomic).
- This is likely also a contributing factor to #1409's residual "handshake" room storm (a losing racer's registry entry being torn down mid-flight can throw inside `flushHeldState`, which is what Engram-obsidian#466 separately hardened) — related but independently scoped fix.

## Test plan
- [x] New test: two concurrent `pushFile` calls for the same file push exactly once (fails pre-fix at 2 calls, passes post-fix)
- [x] `bun test`: 2878 pass / 0 fail
- [x] `tsc -noEmit` + `esbuild` clean
- [x] `biome check` clean
- [ ] Confirm against the actual `test_97`/`98`/`100` E2E suite once this lands — that's the real proof; the unit test only proves the reentrancy gap is closed, not that it's the *sole* cause of the observed doubling

### DIFF
--- a/src/sync.ts
+++ b/src/sync.ts
@@ -3564,6 +3564,19 @@ export class SyncEngine {
 		// server-sanitized-rename check) must be made against the path we
 		// actually pushed, not wherever the file lives by reply time (#245).
 		const pushedPath = file.path;
+		// Re-check the guard HERE, not just at entry (line ~3494): two awaits sit
+		// between that check and this claim (the echo-hash read, acquirePushSlot),
+		// so two overlapping pushFile calls for the same path — e.g. two
+		// overlapping fullSync() sweeps, which nothing serializes — can both pass
+		// the stale check before either claims the slot, and both push. For a
+		// brand-new CRDT note that means two independent genesis lineages landing
+		// on the same server row. Nothing async happens between this check and the
+		// claim below, so this pairing IS atomic (single-threaded JS).
+		if (this.pushing.has(pushedPath)) {
+			this.releasePushSlot();
+			devLog().log("push", `skip (already pushing, lost the race): ${pushedPath}`);
+			return false;
+		}
 		this.pushing.add(pushedPath);
 		this.lastError = "";
 		this.emitStatus();
--- a/tests/sync.test.ts
+++ b/tests/sync.test.ts
@@ -2628,6 +2628,40 @@ describe("push concurrency limit", () => {
 
 		expect(completedCount).toBe(10);
 	});
+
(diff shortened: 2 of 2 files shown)


Rasbandit:
## Correcting this PR's justification before merge

**This did NOT fix the lineage doubling it was written for.** The counts were unchanged with and without it (7/78, 12/135, 12/146 before and after). The real cause was a mis-scoped e2e assertion — the lineage probe sampled a session-shared vault and counted *other* tests' legitimately two-writer notes. Fixed separately in Engram#1459.

I also previously credited this PR with taking `e2e-crdt` from 4 failures to green. That attribution does not hold either: `e2e-crdt` also passed on #466's run, which does not contain this change.

## What it does stand on

A genuine TOCTOU in `pushFile`'s reentrancy guard: `this.pushing.has(path)` is checked at entry, but two `await`s (echo-hash read, `acquirePushSlot`) sit between that check and `this.pushing.add(path)`, so two overlapping calls for the same path can both pass the stale check and both push. The unit test fails pre-fix (2 pushes) and passes post-fix (1).

Note the "drop on losing the race" behaviour is **not new** — the entry guard already returns false for the common case where the second call starts after the first has claimed. This change only closes a narrow window in which the guard silently didn't work.

## CI

First post-rebase run failed `test_82_resumed_device_stale_idmap`; re-run came back **110 passed, 1 skipped, 0 failed**. test_82 passed on both comparison runs too, and the mechanism above can't plausibly cause it, so: flake.

PR #470, opened 2026-08-25
fix(auth): stop offering API keys to tiers that cannot use them

Rasbandit:
Copy only, no behaviour change. Follows Engram#1465 (API keys move to Pro).

Free never could use API keys: `api_rps_cap` has been 0 on Free since launch, and both `RequireApiRpsBudget` and `ChannelGate.api_access/2` reject a zero cap, so a pasted key failed REST and the sync socket alike. The plugin offered the path anyway. Engram#1465 extends that to Starter, so the copy needs to match on both counts.

- **README + user guide** — the guide told anyone whose sign-in window stalled to "fall back to an API key", which is now the one thing Free and Starter cannot do.
- **Settings** — `renderAuthSection` lives in `self-hosted-tab.ts` but `connection-tab.ts` calls it in cloud mode too, so the API key heading and token field render for every cloud user. Cloud now says the key needs Pro and points at sign-in; self-host keeps the plain wording since it enforces no limits.
- **`limit-copy.ts`** — `api_access_not_available` and `api_write_not_available` had no entry and fell through to a generic "Limit reached. Upgrade to continue.", which does not tell the user what to do. Both now name the cause and point at sign-in, which works on every plan.
- Dropped the dead `realtime_disabled` entry. No tier gates real-time sync and the server never emits that reason.

## Verification

`bun test` 2879 pass / 1 skip / 0 fail, `bun run build` clean, biome + `lint:obsidian` + `lint:css` clean.

https://claude.ai/code/session_01KnxMyiSWtoZ8CTmeHjbwoA

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Point your client at the Engram MCP server (`https://mcp.engram.page` on the hos
 
 ## Privacy
 
-- **Network use & accounts.** The plugin talks only to your Engram server, nothing else, no middlemen. You connect with your Engram account via OAuth sign-in or an API key.
+- **Network use & accounts.** The plugin talks only to your Engram server, nothing else, no middlemen. You connect with your Engram account via OAuth sign-in, which works on every plan. API keys are also supported, but on Engram Cloud they require the Pro plan; self-hosted servers have no such limit.
 - **No telemetry.** Optional remote logging (off by default) sends error/lifecycle events only to your server.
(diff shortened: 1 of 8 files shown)


Rasbandit:
**Merge order: this must land after Engram#1465 deploys.**

The settings copy tells cloud users API keys need Pro, which only becomes true once #1465 is in prod. Before that, Starter can still use keys and this text is wrong. Same dependency applies to engram-marketing#180.

The `limit-copy` and `parseLimitExceededError` fixes are independent and safe in any order.

PR #471, opened 2026-08-25
fix(crdt): converge cold notes without opening a room (#1409)

Rasbandit:
## What

The client half of #1409: a diverged **cold** note now converges over the room-free `crdt_doc_state` read instead of a syncStep1 re-handshake.

Pairs with **engram-app/Engram#1467**, which adds the frame. **Land the backend first.** This side falls back to the room handshake when the frame is rejected, so it is not blocked on merge order for correctness — only for the room saving to actually take effect.

## Why

A bulk first sync opened roughly one server room per note. Attribution across all three e2e devices on a 250-note import put **100% of the rooms at one call site**: `applyChange`'s "diverged cold note, socket re-handshake" leg — a note nobody has open in an editor, whose STEP1 the backend routes through `ensure_room`.

## How

That site fires `crdt_doc_state` and applies the result via `applyRemoteUpdate`. Same monotonic merge STEP2's ops would have been, which is what makes it a legal substitute — and what the catch-up feed's plaintext `content` is not (a checkpoint-lagged snapshot *write* can revert a fresher live merge, the D2 stomp class).

Two things this had to get right:

- **`applyRemoteUpdate` does NOT fire `onSynced`** — only an inbound syncStep2 does. Without an explicit `markSynced`, the staged episode never commits, `serverHash` is never recorded, and every later catch-up re-fetches the note forever. This would have shown 0 rooms while silently never converging. The unit test drives the commit *through the port* rather than calling `commitCrdtConvergence` by hand, so a version that never triggers it fails.
- **Any failure falls back to the room handshake** — an old backend that rejects the unknown frame, a rate limit, a dropped socket. Convergence is the invariant; the room is only the expensive way to reach it.

`stageConvergence` is split out of `stageAndConverge` so both delivery paths share ONE writer of the staged episode. That helper exists precisely because the stage was hand-copied five times and forked.

## Verification

Measured on the local CRDT stack, 250-note import: **3-4 handshake rooms before, 0 after.** `test_77` passes at `HANDSHAKE_ROOM_RATCHET = 0`.

- 4 new unit tests, mutation-checked (disabling the room-free path fails them)
- 2879 existing plugin tests pass
- 17/21 `tests/crdt/` e2e pass locally. The 4 failures are the known headless-Chromium `Locator.click` class, **confirmed by control run**: the same 4 fail identically against `origin/main` on the same stack. Checked because `test_web_edit_reaches_obsidian_that_missed_room_open` sits squarely in this change's blast radius.

Refs #1409

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -572,6 +572,31 @@ export class NoteChannel {
 		};
 	}
 
+	/** Full Yjs state for `docId`, read WITHOUT starting a server room (#1409).
+	 *
+	 *  The alternative — a syncStep1 `crdt_msg` — routes through the backend's
+	 *  `ensure_room`, so every cold note that needed converging allocated a room
+	 *  for a note nobody has open. The server answers this one off the persisted
+	 *  snapshot + update-log tail (`CrdtTransport.read_delta`), which spawns
+	 *  nothing.
+	 *
+	 *  Returns base64 Yjs v1 update bytes. Applying them is a MERGE, so a
+	 *  checkpoint-lagged snapshot converges without reverting anything — the
+	 *  property that makes this a legal substitute for STEP2, and that the
+	 *  catch-up feed's plaintext `content` snapshot does NOT have.
+	 *
+	 *  Rejects carry the same reasons as a handshake: `rate_limited` (retryable),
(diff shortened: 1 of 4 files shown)


Rasbandit:

PR #472, opened 2026-08-25
diag(crdt): log which genesis branch each create takes (#1409) — DO NOT MERGE

Rasbandit:
**DIAGNOSTIC ONLY — do not merge.** Opened solely to get a `pr-build` prerelease for BRAT.

## Evidence so far

A real 423-item Obsidian first sync (single device, nothing open in an editor):

| | |
|---|---|
| Notes | 317 |
| Rooms opened | **225** (0.71/note) |
| Source | `edit` × 225; `handshake` 0, `create_batch` 0 |
| `crdt_update_log` rows written | **0** |
| Notes with server-side `crdt_state` | 317/317 |

Every note already had its full state server-side from the roomless genesis
seed in `crdt_create`. Those 225 rooms opened, applied nothing, and drained.

## Mechanism

The seeded fast path (`sync.ts`, both `applyCrdtCreateAck` and `pushFile`)
applies the same bytes locally with provider origin — its own comment says it
"does NOT re-broadcast and does NOT open a room". It's gated on
`seeded && sameId && genesis && !effectiveIdHasHistory`. When any condition
fails, the `else` leg runs `routeModify`, whose diff emits a local update, the
provider broadcasts it as a `sync_update`, and that frame classifies `edit` →
`ensure_room` → room. ~71% of notes take the else leg.

## What this build adds

One `GENESIS_BRANCH` log line at each of the two guards, recording FAST/SLOW
plus all four conditions (`seeded`, `sameId`, `genesis`, `throwaway`,
`hasHistory`). The next real import answers which condition fails, instead of
another hypothesis.

Branched off `fix/1409-roomless-cold-note-pull` (#471) so the fix PR stays clean.

Refs #1409

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -572,6 +572,31 @@ export class NoteChannel {
 		};
 	}
 
+	/** Full Yjs state for `docId`, read WITHOUT starting a server room (#1409).
+	 *
+	 *  The alternative — a syncStep1 `crdt_msg` — routes through the backend's
+	 *  `ensure_room`, so every cold note that needed converging allocated a room
+	 *  for a note nobody has open. The server answers this one off the persisted
+	 *  snapshot + update-log tail (`CrdtTransport.read_delta`), which spawns
+	 *  nothing.
+	 *
+	 *  Returns base64 Yjs v1 update bytes. Applying them is a MERGE, so a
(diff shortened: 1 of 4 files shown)


Rasbandit:

PR #473, opened 2026-08-25
fix(crdt): don't open a room per note when the server re-ids a create (#1409 push half)

Rasbandit:
The **push half** of #1409. Stacked on #472 (diagnostic) → #471 (receive-half fix); still carries the `GENESIS_BRANCH` log line for verification, which comes out before this is merge-ready.

## Measured, real 423-item Obsidian first sync, single device, nothing open in an editor

| | |
|---|---|
| Notes | 317 |
| Rooms | **225** (0.71/note), `edit` × 225, `handshake` 0, `create_batch` 0 |
| `crdt_update_log` rows those rooms wrote | **0** |
| Notes already holding server-side state | 317/317 |

The rooms did no work. Every note's state was already server-side from the roomless genesis seed.

## Cause

Instrumented the branch point — **317/317 creates took the SLOW leg**, all `seeded=true sameId=false hasHistory=false`. Room count (225) equals the `throwaway=false` subset exactly.

The plugin proposes note ids belonging to a **different vault**, confirmed against the DB:

```
local id vault → server id vault
c15d1c9e ("no room v2")  →  9e3dab50 (v3)   same_vault: false
18037479 (older vault)   →  dcde98c2 (v1)   same_vault: false
```

The server can't reuse a foreign-vault id, returns a fresh one and seeds our bytes into it, `sameId` goes false, the fast path is skipped, `routeModify` diffs the body onto the new id, and the provider broadcasts a `sync_update` → edit-class frame → `ensure_room` → one room per note.

This is why the e2e harness reads 0 on the pushing device: it mints brand-new ids every run, so `sameId` is always true there. The bug needs an install that has synced this content before — re-adding a vault, renaming one, or testing repeatedly.

## Why dropping `sameId` is safe

`seeded: true` already proves our genesis bytes landed in an **empty** row:

- the genuine `{:adopted, note}` arm hardcodes `seeded: false`
- `seed_against/7` writes only on its empty-row clause; the catch-all ("row already carries a different body") declines

`!effectiveIdHasHistory` still carries the anti-doubling weight, and it's checked against the id actually written. Both fast paths already applied to `effectiveId`, so nothing needed retargeting.

## Not fixed here

The underlying cross-vault id reuse — same class as #1318. This stops that reuse from costing a room per note; it doesn't stop the reuse.

## Verification

2883 plugin unit tests pass. Real-vault re-import pending on this build.

Refs #1409

### DIFF
--- a/src/channel.ts
+++ b/src/channel.ts
@@ -572,6 +572,31 @@ export class NoteChannel {
 		};
 	}
 
+	/** Full Yjs state for `docId`, read WITHOUT starting a server room (#1409).
+	 *
+	 *  The alternative — a syncStep1 `crdt_msg` — routes through the backend's
+	 *  `ensure_room`, so every cold note that needed converging allocated a room
+	 *  for a note nobody has open. The server answers this one off the persisted
+	 *  snapshot + update-log tail (`CrdtTransport.read_delta`), which spawns
+	 *  nothing.
+	 *
+	 *  Returns base64 Yjs v1 update bytes. Applying them is a MERGE, so a
+	 *  checkpoint-lagged snapshot converges without reverting anything — the
+	 *  property that makes this a legal substitute for STEP2, and that the
+	 *  catch-up feed's plaintext `content` snapshot does NOT have.
+	 *
+	 *  Rejects carry the same reasons as a handshake: `rate_limited` (retryable),
+	 *  `note_not_found` (the caller's id-map is stale), `bad_doc_id` (terminal).
+	 *  Callers fall back to the room re-handshake — including on an
+	 *  `unmatched_topic`-shaped reject from a backend too old to know the
+	 *  frame, so a new plugin against an old server still converges. */
+	async crdtDocState(docId: string): Promise<{ b64: string; head: string }> {
+		return (await this.sendRequest("crdt_doc_state", { doc_id: docId })) as {
+			b64: string;
+			head: string;
+		};
+	}
+
 	/** Seq-ordered op-log page after `cursorSeq` (single-path catch-up). Each op
 	 *  carries FULL content (a SyncNoteChange or SyncAttachmentChange — the
 	 *  merged notes+attachments feed), so it is causally complete and can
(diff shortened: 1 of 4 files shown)


Rasbandit:
## ⚠️ REGRESSION — do not merge as-is

Real-vault verification on a 423-item import (single device, nothing open in an editor). Same vault content, same machine, four consecutive runs against the same backend:

| Run | Notes | Rooms | Rooms/note | Drift-copies | Conflict files |
|---|---|---|---|---|---|
| v1 | 317 | 212 | 0.67 | 0 | 0 |
| v2 | 317 | 225 | 0.71 | 0 | 0 |
| v3 | 317 | 225 | 0.71 | 0 | 0 |
| **v4 — this PR** | 317 | **23** | **0.073** | **11** | **11** |

The room fix works — ~10× fewer, and the remaining 23 split `edit: 16 / handshake: 7`, a different and much smaller tail.

But it introduces **11 conflict copies** where three prior runs produced zero. That's not an acceptable trade.

### Why the safety argument was incomplete

I verified that `seeded: true` implies the server wrote our bytes into an **empty** row (the real `{:adopted, _}` arm hardcodes `seeded: false`; `seed_against/7` writes only on its empty-row clause), and that `!effectiveIdHasHistory` rules out a rival local lineage. Both hold.

What I never checked is whether the bytes are **current**.

`throwaway=false` — which is exactly the 225 notes that newly take the fast path — means the genesis bytes are encoded from the pre-existing local Y.Doc under the **stale previous-vault id**. That doc carries the *old vault's* content while disk carries the current content. The old SLOW leg ran `routeModify`, which always diffs **current disk**; the fast path applies frozen bytes and then re-merges drift, and that difference is where the conflicts come from.

Measured: v4's branch split is identical to v3 (225 `throwaway=false`, 92 `throwaway=true`), and 11 of the 225 that flipped to the fast path conflicted — ~5%.

### Conclusion

This fix's correctness depends on "the local doc's lineage represents this note's content", and cross-vault id reuse is precisely the case where that is false. It cannot be made safe on its own — it is a symptom fix for a root cause (#1318-class cross-vault id reuse) that has to be fixed directly.

The right fix is to vault-scope the note-id map so the plugin stops proposing another vault's ids at all. Then `sameId` is naturally true, the **existing** fast path fires unmodified, rooms drop, *and* the genesis bytes are guaranteed to be this vault's own.

Superseded by that work. Keeping this open only as the evidence record.

Refs #1409

Rasbandit:

PR #487, opened 2026-09-01
fix(crdt): table-cell editor corruption + trim redundant sync status surfaces

Rasbandit:
## 1. Table-cell editor corruption (prod, reproducible)

Clicking a table cell in Live Preview inserted the **entire note body** into that
cell, froze Obsidian, and synced the corruption to the server.

`registerEditorExtension` installs the live binding into **every** CM6
`EditorView` Obsidian builds. The Live Preview table widget builds a nested
`EditorView` per cell (`editor.tableCell.cm`) and constructs it with the
**parent editor's `owner`** - so `editorInfoField` inside a cell resolves to the
same `MarkdownView` and the same `file.path`.

`editorPath()` keyed only off that path, so the cell's view bound to the whole
note's `Y.Text`. The initial reconcile then read "editor is stale disk, doc is
authoritative", took the `adopt` branch, and painted the entire document into the
cell - which Obsidian's table code round-tripped into the real file.

`ownedMarkdownPath()` binds only when the `EditorView` **is** the owner's own
editor (`info.editor.cm === view`). Same guard Relay's `findView` has always had,
which is why Relay never had this bug. Covers every nested editor built off a
parent owner, not just table cells.

Note the check is null at construction time (the `ViewPlugin` constructor runs
inside `new EditorView(...)`, before Obsidian assigns `owner.editor.cm`);
`update()` re-resolves and `needsReattach` fires on the null -> path transition,
so the real editor binds on its first `ViewUpdate`.

Prod evidence: `30 Growth/Launch/Video 1 - Beta Invite Intro (script).md` has six
table rows each grown to ~17,900 chars holding the whole body, newlines
serialized as `<br>`. **That file still needs repairing - this PR only stops it
happening again.**

Context doc: `docs/context/live-binding-table-cell-editor.md`.

## 2. Advanced tab: debounce control removed

The 2s modify debounce is an internal sync tuning knob, not a preference worth
exposing. `debounceMs` stays in settings at its 2000 default (sync.ts still reads
it, ~60 tests still set it) - only the text field and its now-empty "Sync
behavior" heading are gone.

## 3. Sync Center: duplicate status readouts removed

The tab showed connection state twice - the settings status strip at the top of
the pane, and the Sync Center header's own dot + `Engram Sync - <state>` line.
The header is badges only now (plan-skip / needs-attention / retrying / ignored /
queued-reason), and it does not render at all when there is nothing to badge.

Same reasoning drops five live counters from Stats: Last sync, Live (WebSocket),
Pending in queue, Issues, Ignored. Each is on the status strip or has its own
section directly above it. Stats is static vault facts now: Local notes, Local
attachments, Vault, Vault ID.

## Verification

- `bun run build` clean (tsc + esbuild)
- `bun test` 3012 pass, 1 skip, 0 fail
- biome clean
- New coverage: `tests/crdt-live-binding-decisions.test.ts` -> `describe("ownedMarkdownPath")`, 5 cases (own editor, inherited cell info, missing file, non-markdown, unassigned `cm`)

https://claude.ai/code/session_01NTroEXqCfFBz6YyEtH5bEC

### DIFF
--- a/docs/context/live-binding-table-cell-editor.md
+++ b/docs/context/live-binding-table-cell-editor.md
@@ -1,0 +1,69 @@
+# Context Doc: Live-binding attached to the table-cell editor — whole document adopted into one cell
+
+_Last verified: 2026-08-31_
+
+## Status
+Fixed (PR #487)
+
+## Symptom
+Click into a table cell in Live Preview. The entire note body is inserted into
+that cell, Obsidian freezes, and the corruption is written to disk and synced to
+the server. Reliably reproducible; hit in production on
+`30 Growth/Launch/Video 1 - Beta Invite Intro (script).md`, where six table rows
+each grew to ~17,900 characters holding the whole body with newlines serialized
+as `<br>`.
+
+## Mechanism
+`this.registerEditorExtension([liveBindingPlugin])` (`src/main.ts`) installs the
+binding into **every** CM6 `EditorView` Obsidian builds — not just the leaf's
+markdown editor.
+
+Obsidian's Live Preview table widget builds a nested `EditorView` per cell
+(`editor.tableCell.cm` in `app.js`) and constructs it with the **parent editor's
+`owner`**. `editorInfoField` inside a cell therefore resolves to the very same
+`MarkdownView`, carrying the very same `file.path`.
+
+The old `editorPath()` keyed only off that path, so the cell's `EditorView` bound
+to the whole note's `Y.Text`:
+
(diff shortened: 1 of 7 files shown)


Rasbandit:
E2E proof paired in engram-app/Engram#1537 (same branch name).

`e2e/tests/crdt/test_table_cell_live_binding.py` — **RED** on the commit before the fix (viewer #2, cell editor holding 27,823 chars of the note), **GREEN** on this fix, 3/3 repeat runs.

Issue #339, opened 2026-07-27
hasServerNote stays false after a note heals via STEP2, so its next push takes the genesis path


Rasbandit:
Follow-up from the #1130 / test_48 work (#335, #337). Not urgent, no data loss — filing it because the decision is a real one and shouldn't be made incidentally inside an unrelated PR.

## The gap

`SyncEngine.hasServerNote` (`src/sync.ts:1854`) is the oracle for "the server holds a row for this note":

```ts
hasServerNote(noteId: string | null): boolean {
	if (!noteId) return false;
	const path = this.noteIdMap?.pathForId(noteId);
	if (!path) return false;
	return this.getCrdtHead(path) != null;
}
```

`crdtHead` is written by exactly two things: applying a `note_yjs_update` fan-out, or a local create-ack (the `CRDT_HEAD_CREATED` sentinel). A **completed STEP1/STEP2 room handshake sets nothing** — `commitCrdtConvergence` records `serverHash`/`version`/`seq`, never a head.

So after the #1130 heal (`socketConverge` → `reset`+`enroll` → the server answers with the note's full state), the device demonstrably knows the server has this note — it just merged the server's own ops for it — and the oracle still says false.

## What that costs

On the note's next local edit, `pushFile`:

- skips the live-CRDT branch (`src/sync.ts:2550`, gated on `hasServerNote`)
- takes the genesis branch instead (`src/sync.ts:2667`, gated on `!hasServerNote`) and sends a `crdt_create` for a note that already has a row

The server's idempotent same-path arm (`engram` `lib/engram/notes.ex:687-689`) returns the existing row and the ack sets `CRDT_HEAD_CREATED`, so it self-corrects after one save. `splitGenesisVsKnown` (`src/sync.ts:6145`) misclassifies the same notes on a bulk push, routing them through `crdtCreateBatch`.

**This is safe, and I checked the part that could have made it unsafe:** the genesis create carries a full-state frame, but `seed_genesis_if_empty` only applies it when the room doc is empty. A healed note's doc is non-empty, so the create returns `:skipped` and the body cannot double (the #846 guard). Confirmed against `create_one` in `crdt_channel.ex`.

So the actual cost is a redundant `crdt_create` round-trip per healed note on first edit, plus the note's ops sitting held behind `canSendLive` until that ack lands. Wasted work and a misleading `route:` log line, not corruption.

## Options

[...]

PR #497, opened 2026-09-01
fix(crdt): re-drive the create for an orphaned index claim

Rasbandit:
Fixes the client half of engram-app/engram#1550. Found while checking #1401's prerequisite; it is a live prod data-integrity bug, not a theoretical one.

## The state

`getOrMint` publishes a claim into `filemeta_v0` the moment it mints, without waiting for `crdt_create` to be acked. When the create then fails to land, the claim stays:

- `filemeta_v0` holds `path → note_id`
- no server row exists for that id
- the file is still on disk

The note exists on **exactly one device**. Every `crdt_msg` for it is dropped `note_not_found`, and the server's `ProjectVaultIndex` retries the unresolvable entry forever — `applied=0 unresolved=N` for that whole vault, permanently, which also destroys the one signal we have for index/row disagreement.

## Measured

Prod, 2026-09-01. Six ids took `note_not_found` in one 8-minute window on one vault:

```
00:11:16  01a05a4e-4c68…      00:12:09  01a05a4f-1cf8…  ← stranded
00:11:37  01a05a4e-9ebb…      00:18:43  01a05a55-2049…
00:11:47  01a05a4e-c682…      00:19:48  01a05a56-1f8e…  ← stranded
```

Four healed. The two that did not are exactly the two whose claim had already landed. Both uuid7 ids decode to a mint timestamp equal to the second of their first rejected `crdt_msg` — ops went out before any create could round-trip. They stayed stranded for **16+ hours** with no user-visible signal.

## Why the existing heal could not fire

`ensureNoteIdMapped` early-returns on `pathForId(noteId) !== null`. For these ids the mapping **is** the orphaned claim — the id vouches for itself, so the guard reads this device's own unconfirmed assertion as proof.

Reaching the reconcile would not have helped either: that sweep iterates the notes the **server** lists, so an id the server has never had appears nowhere in it. It would cost a whole-vault manifest fetch and change nothing.

I tried the obvious-looking fix first — also demanding `hasServerNote` on that early return — and reverted it. It buys nothing the repair below does not already cover, breaks the existing "a known id is a no-op (no manifest fetch)" test, and costs a manifest fetch per unconfirmed id.

## The fix

This is a create that went missing, not a mapping that drifted, and the repair for a missing create is to make it again. `repairUnconfirmedClaim` re-enqueues the durable `crdt_create` on the same `note_not_found` signal, using the same call `pushFile` already makes for a note with no server row.

- **Same id, not a fresh mint.** The local Y.Doc holding the user's content is keyed by it and the claim already names it, so one create makes all three identity sources agree at once and carries the body up through the queue's genesis frame. A re-mint would strand that doc and leave a second claim to clean up.
- **Gated on `hasServerNote`**, which reads `crdtHead` — written only by a create-ack or a server-delivered head, so it can never be satisfied by this device's own claim.
- **No local file → not our business.** Nothing to create; that is ordinary drift and stays with the reconcile.
- **Idempotent.** `CrdtOpQueue` keys by `docId` and supersedes in place, so a create already queued is replaced, never duplicated.

It also **self-heals vaults already in this state** on the next run, which is how the two stranded notes reach the server without anyone touching prod data.

## Not covered here

The underlying ordering — claim published before create-ack — is untouched. `SyncStore` already has `pendingUpload`/`confirmUpload` built for exactly that gate, with **zero production callers**; wiring it is the real prevention and belongs in its own PR against #1550, because holding claims back until confirmation changes `commit()`'s all-or-nothing folder-rename semantics. This PR makes the state recoverable; it does not stop it arising.

## Test

Three cases in `tests/sync-id-adoption.test.ts`: the create is re-enqueued under the same id and the manifest sweep is skipped; a note the server does know is left alone; a claim with no file behind it is not re-created.

Mutation-checked — commenting out the `repairUnconfirmedClaim` call turns the first test red, so it is not passing vacuously.

Full suite 3030 pass / 0 fail; `tsc`, `biome check`, `bun run build` clean.

Refs engram-app/engram#1550, engram-app/engram#1401

### DIFF
--- a/src/crdt/wiring.ts
+++ b/src/crdt/wiring.ts
@@ -19,6 +19,7 @@ type WiringSyncEngine = Pick<
 	| "reconcileNoteIdMapFromManifest"
 	| "isSyncBlocked"
 	| "ensureNoteIdMapped"
+	| "repairOrphanedClaim"
 	| "clearPushedBaselineForId"
 	| "applyPushedNoteUpdate"
 	| "discoverAnnouncedNote"
@@ -497,6 +498,12 @@ export function createCrdtWiring(deps: CrdtWiringDeps): CrdtWiring {
 		// Un-bank first — the reconcile below can re-key the map out from under
 		// the lookup this needs.
 		syncEngine.clearPushedBaselineForId(docId);
+		// THIS reply — not the announce — is the server stating it has no row for
+		// the id, which is what licenses re-driving the create (#1550). Handled
+		// before the reconcile because that sweep cannot see an orphan at all:
+		// it iterates the notes the server LISTS, and an id the server has never
+		// had appears in none of them.
+		if (syncEngine.repairOrphanedClaim(docId)) return;
 		syncEngine.ensureNoteIdMapped(docId);
 	};
 
--- a/src/sync.ts
+++ b/src/sync.ts
@@ -1112,6 +1112,18 @@ export class SyncEngine {
 		// boundary — `reconcileNoteIdMapFromManifest` has three other callers
 		// and carries its own tombstone skip; see there.
 		if (this.recentlyDeleted.has(noteId)) return;
+
+		// An orphaned claim is repaired HERE, not by the reconcile below: the
+		// manifest cannot explain an id the server has never had, so the sweep
+		// would fetch the whole vault to learn nothing. See the method.
+		if (this.repairUnconfirmedClaim(noteId)) return;
+
+		// Already mapped — ordinary drift is repaired by the reconcile, and the
+		// ORPHANED-claim case (mapped by a claim the server never honoured) was
+		// handled above, because the manifest cannot explain an id the server
+		// has never had. Weakening this to also demand `hasServerNote` was tried
+		// and reverted: it buys nothing the repair above does not already cover
+		// and costs a whole-vault manifest fetch per unconfirmed id.
 		if (this.noteIdMap.pathForId(noteId) !== null) return; // already mapped
 		if (this.idMapReconcileInflight) {
 			this.idMapReconcileQueued = true;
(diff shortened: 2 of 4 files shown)


Rasbandit:
Review pass applied (`c8b2fdc`). Three findings were real; recording what changed and why.

**1. The gate was weaker than its own comment claimed — fixed by moving the trigger.** `hasServerNote` false does *not* mean the server lacks the row. `renamePath` deliberately drops `crdtHead` at the new path on every rename so `pushFile` takes the genesis branch, and the converge legs only mark a note server-known when staged content is non-empty (#339 — an empty converged doc is not proof). Since `ensureNoteIdMapped` is also called from the `crdt_doc_ready` announce, an empty note another device opens — the exact case `wiring.ts:466` exists for — would have fired a spurious create on every announce, as would any note inside the window after a rename.

Rather than strengthen the gate, the repair moved to the `note_not_found` reply: the one place the **server** states it has no row for the id. The announce path no longer reaches it at all. That is a better fix than a bigger conjunction, because it stops guessing from a proxy and uses the authoritative signal.

**2. Missing eligibility/size gate — added.** An over-cap (>4 MB) or ineligible note is kept off the CRDT path by design and therefore never carries a `crdtHead`, so it satisfied every other condition. The queue would have built an over-cap genesis frame and burned all 8 attempts before surfacing a user-visible drop. Now gated on `isCrdtEligiblePath` + the cap, matching every sibling `crdtEnqueue({kind:"create"})` site. `stat.size` is the same UTF-8 byte count `exceedsCrdtNoteLimit` measures, without reading the file.

**3. Unbounded warn/enqueue — now once per id per session.** The durable queue owns retrying a failed create; re-enqueuing on every later `note_not_found` adds only a Loki warn per reconnect.

**4. Scope gap — documented rather than built.** The repair needs a `crdt_msg` to escape, and `canSendLive` holds ordinary ops behind `hasServerNote`, so the note must be enrolled: open in an editor, or in `unsentDocIds` and re-STEP1'd by `reEnrollUnsent` on rejoin. I checked the prod logs against this — the stranded ids reappear across multiple task ARNs, i.e. on every reconnect, so the real case *is* reached. An orphan never opened and never edited is not, and the doc comment now says so.

### On the tests

The review was right that the important one was missing, and worse than it looked: **the wiring call site had no coverage at all** — deleting `repairOrphanedClaim` from `onCrdtNoteNotFound` left the entire 3000-test suite green. Same shape as the rate-limit mutation this repo has been bitten by before. Added a wiring test that pins it and verified it goes red when the call is removed, plus over-cap and ineligible-path cases.

One correction to my own new test: I wrote the ineligible case as a `.canvas` file. Canvas **is** CRDT-eligible (`isCrdtEligiblePath` = markdown OR canvas), so that was a note the repair *should* cover. Switched to an attachment.

Full suite 3034 pass / 0 fail; `tsc`, `biome`, build clean.

### Still owed

An e2e. Everything here is unit-level and asserts the create is *enqueued*, not that the stranded note reaches the server. `e2e/tests/crdt/test_index_crdt_client_writes.py` already has the right oracle (reads `filemeta_v0` back through `backend_rpc` + `load_doc/2`), and the state can be constructed deterministically — delete the note **row** via `backend_rpc` while leaving the claim — with no fault injection needed. Not in this PR.

Rasbandit:
**e2e proof landed** (engram#1551), run both ways:

| plugin ref | result |
|---|---|
| this PR (`c8b2fdca`) | `28 passed` — the orphan e2e **passes** |
| plugin `main` (`05d3223`) | `1 failed, 27 passed` — it **fails**, on `note row ... did not reappear within 120.0s` |

So the stranded note is reproduced without this fix and recovered with it, under the same id. That closes the gap I flagged earlier: the unit tests assert the create is enqueued, this asserts it actually lands.

PR #520, opened 2026-09-18
feat(i18n): translate Notice text into 10 languages

Rasbandit:
Follows #519 (README translations). This is the first UI surface, not the whole plugin.

## What

- `src/i18n/index.ts`: a `t()` helper plus locale dictionaries for zh, zh-TW, ja, ko, de, fr, es, pt, ru, it.
- All 38 literal `new Notice(...)` strings wrapped in `t()`, across 9 files.
- `src/i18n/locale/README.md` for whoever adds the next language.

Language follows Obsidian's own setting via `getLanguage()`. No dependency added; the whole `src/i18n` tree is 56K of source.

## Design, and why it differs from the reference plugins

**The English string is the key, and there is no `en.ts`.** A missing translation falls through to a readable English sentence rather than a bare identifier. Editing Toolbar (1.91M downloads) and Excalidraw (8.04M) both hand-maintain an identity map for English, and Editing Toolbar's has already drifted from itself: one entry reads `'Choose between fixed position or cursor following mode.': 'Choose between fixed position, cursor following or Top mode.'`. Not shipping the file removes the failure mode.

**Locale files use Obsidian's own codes.** `zh`, `zh-TW`, `pt-BR`; there is no `zh-CN`. Excalidraw and remotely-save both read `moment.locale()`, get `zh-cn`, and need a normalization layer (Excalidraw carries an explicit `if (lang === "zh") lang = "zh-cn"` for its issue #2247). Reading `getLanguage()` and naming files to match removes that layer. A regional code falls back to its base language, so `en-GB` reaches `en`.

**Counted strings carry one template per CLDR plural category,** selected by native `Intl.PluralRules`. Two sites previously hardcoded English plural rules inline:

```ts
const noun = count === 1 ? "file" : "files";
new Notice(`Engram: ${count} ${noun} failed to sync${detail} — open Sync Center`, 10_000);
```

No flat dictionary can express that. Russian needs `one`/`few`/`many`; zh, ja and ko need a single plain string.

**`getLanguage()` is guarded.** It may postdate our `minAppVersion` of 1.7.2, so `typeof getLanguage !== "function"` yields English rather than throwing. I did not verify which app version introduced it.

## Verification

```
3170 pass, 1 skip, 0 fail   (31 of those new)
tsc -noEmit   clean
biome check   clean
eslint        0 warnings
bun run build succeeds
```

`tests/i18n-keys.test.ts` is the anti-drift guard: it fails if any locale holds a key that no `t()` call site uses, which is what catches a typo'd key or English copy edited without its translations.

## Scope left out on purpose

- Internal log and error strings. Of 878 prose literals in `src/`, only ~188 are user-facing; `src/sync.ts` alone holds 295 log lines nobody sees.
- The 20 `new Notice(...)` sites built from composed fragments rather than one literal.
- The other surfaces: settings tabs (~68 strings), the sync preview and progress modals, Sync Center, search. One PR each.
- No per-plugin language override setting. Obsidian's setting is the only input. If we add one, avoid remotely-save's bug, where an explicitly chosen locale falls into the `else` branch and silently returns English.

## Known risk

Nobody on the team reads zh, ja, ko or ru, so translation errors will ship silently. That is an accepted trade here: a rough translation beats the wrong language.

https://claude.ai/code/session_01LnWo9H55uN6g9cTe3Mafnh


Rasbandit:
Folded into #525, which now targets main directly and carries all four commits.

PR #524, opened 2026-09-18
feat(i18n): route every user-facing surface through t()

Rasbandit:
**Stacked on #520.** Base is `feat/plugin-i18n`, so review that one first.

Takes `t()` call sites from 39 to **204**, covering every user-facing surface: settings tabs, the sync preview / progress / device-flow / sync-log / upgrade modals, Sync Center, search, the issue-store titles, and the status bar.

Locale dictionaries are **unchanged here**. Every newly wrapped string still renders in English, so this PR is behaviour-neutral by design and reviewable as pure mechanism. Translation content lands per batch after it.

## Three things needed restructuring, not wrapping

**`tInto()` for sentences built around a styled child.** This was three literals:

```ts
matchRow.createSpan({ cls: "...-match-label", text: "Your vault shares " });
matchRow.createSpan({ cls: "...-match-value", text: `${match}%` });
matchRow.createSpan({ cls: "...-match-label", text: " of its data with Engram" });
```

Two of those three are ungrammatical alone and the slot cannot move, which most languages need it to. It is now one key, `"Your vault shares {percent} of its data with Engram"`, with the styled span built at the slot. Same treatment for `"Type {keyword} to confirm:"`. A translation that omits the slot still renders as plain text.

**`upgrade-required.ts` appended a clause to a sentence:** `` `...too old to sync${needs}.` `` where `needs` was `" (needs X or newer)"`. Now two complete alternative sentences, so a translator can place the clause.

**Bare English in module-level maps.** `QUEUED_REASON_TEXT` and advanced-tab's detected-folder descriptions were reached only through a variable, so `t(var)` would have made them invisible to the orphan test. Their values are wrapped in place instead, making them real scannable keys.

## Verification

```
3170 pass, 1 skip, 0 fail
tsc -noEmit   exit 0
biome check   exit 0
eslint        exit 0 (0 warnings)
bun run build exit 0
```

## Known remaining tail

The call-site scan cannot see copy returned from a pure helper, and there is more of it than my earlier estimate implied:

- `sync-preview-modal.ts`: `planLoadErrorMessage()`, `describeCreateVaultError()`, `skippedAttachmentsLine()`
- `limit-copy.ts`: `toastFor()`
- `sync-plan-format.ts`: `plural()` / `pluralWord()` and the option-breakdown copy

Those return English strings to already-wrapped call sites, so they read as translated surfaces that are still English. They also carry their own English plural logic, so they want the same `Intl.PluralRules` treatment.

Also unwrapped: three label fragments in advanced-tab (`"Source: "`, `"Version: "`, `"License: MIT"`), which are concatenated with a value and need the `tInto` or placeholder treatment.

Deliberately untouched: internal log lines, server error text, and syntax examples such as the ignore-pattern placeholder.

https://claude.ai/code/session_01LnWo9H55uN6g9cTe3Mafnh


Rasbandit:
Folded into #525, which now targets main directly and carries all four commits.

